PR #69, opened 2021-03-11
Add featured snippet (document search) support

cea2aj:
Add support for a new direct answer type: a featured snippet

The direct answer gets a new field: the type field which indicates the type of direct answer. A `DirectAnswerType` enum describes these possible types. I chose to use strings rather than the enums on the models because otherwise the generated docs would create a link to a `DirectAnswerType.FieldValue` or a `DirectAnswerType.FeaturedSnippet` which don't exist in the docs. End users however can use the enums to check the direct answer type.

The `DirectAnswer` models loses the fields specific to the 'FIELD_VALUE' direct answer. Two new DA types are created: `FeaturedSnippetDirectAnswer` and `FieldValueDirectAnswer`. These both extend `DirectAnswer`. A `Snippet` models is created which contains the text and the location of where the direct answer was found.

J=SLAP-1125
TEST=manual, auto

Use the testing library and execute universal searches which contain featured snippet direct answers and field value direct answers. Confirm that the models produced by core match the public interfaces. Add unit tests.


src/models/searchservice/response/DirectAnswer.ts
```diff
@@ -1,23 +1,61 @@
...
+}
+
+/**
+ * A direct answer which was found within a document.
+ *
+ * @public
+ */
+export interface FeaturedSnippetDirectAnswer extends DirectAnswer {
+  /** {@link DirectAnswerType}.FeaturedSnippet. */
+  type: 'FEATURED_SNIPPET';
```
cea2aj, line 28: It's possible to use the enum directly here (e.g. DirectAnswerType.FeaturedSnippet), however that causes the documentation generation to creates a link to a page that doesn't exist. This is a small workaround which allows the enum to be used for checking the type without causing  a broken link to be generated.

src/models/searchservice/response/DirectAnswer.ts
```diff
@@ -1,23 +1,61 @@
...
   verticalKey: string;
-  /** The name of the entity of the direct answer. */
+}
+
+/**
+ * A direct answer which was found within a document.
+ *
+ * @public
+ */
+export interface FeaturedSnippetDirectAnswer extends DirectAnswer {
```
tmeyer2115, line 26: Can we move `FeaturedSnippetDirectAnswer` and `FieldValueDirectAnswer` to their own files?

src/models/searchservice/response/DirectAnswer.ts
tmeyer2115: Is this due to a known bug with TSDoc?

src/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -1,14 +1,28 @@
...
-  return {
-    relatedResult: ResultsFactory.fromDirectAnswer(data.relatedItem.data),
-    verticalKey: data.relatedItem.verticalConfigId,
-    entityName: data.answer.entityName,
-    fieldName: data.answer.fieldName,
-    fieldApiName: data.answer.fieldApiName,
-    value: data.answer.value,
-    fieldType: data.answer.fieldType
-  };
+export function createDirectAnswer(data: any): FeaturedSnippetDirectAnswer | FieldValueDirectAnswer | undefined {
```
tmeyer2115, line 5: It feels a bit funny that this function can return `undefined`. Could we throw an error instead if we don't recognize the DA type?

cea2aj: sure

src/models/searchservice/response/DirectAnswer.ts
cea2aj: It's actually an issue with api-extractor. I'll message the devs and see if it's a bug

src/models/searchservice/response/DirectAnswer.ts
```diff
@@ -1,23 +1,61 @@
...
+import { Snippet } from './Snippet';
 
 /**
  * A direct answer to a search.
  *
  * @public
  */
 export interface DirectAnswer {
+  /** The {@link DirectAnswerType}. */
+  type: DirectAnswerType | string;
```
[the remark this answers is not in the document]
cea2aj: Since I'm using strings to describe the type on the `FeaturedSnippetDirectAnswer` and the `FieldValueDirectAnswer`, it needs to allow strings

src/models/searchservice/response/DirectAnswer.ts
cea2aj: sure

src/transformers/searchservice/createDirectAnswer.ts
creotutar, line 5: Why do we allow `undefined` here? I would expect some indication that we didn't create a direct answer right?

src/models/searchservice/response/DirectAnswer.ts
creotutar: Do we want to form our code around documentation tools? I would expect this to be only of type `DirectAnswerType`, as it's easier to reason what values can go into the type

cea2aj: I wanted to avoid having broken links in our documentation, however it might be fine if we put a working link right next to it

src/transformers/searchservice/createDirectAnswer.ts
cea2aj: A direct answer is an optional field of a `UniversalSearchResponse`, so returning undefined is what we want to do if the backend doesn't return a direct answer. Per Tom's suggestion, I will update this function to throw an error if we get an unknown DA type

src/models/searchservice/response/FeaturedSnippetDirectAnswer.ts
```diff
@@ -0,0 +1,22 @@
...
+import { DirectAnswerType } from './DirectAnswerType';
+import { Result } from './Result';
+import { Snippet } from './Snippet';
+
+/**
+ * A direct answer which was found within a document.
+ *
+ * @public
+ */
+export interface FeaturedSnippetDirectAnswer extends DirectAnswer {
```
oshi97, line 11: do we need/want this extends? It looks like both of the child interfaces have type, value, relatedResult, and verticalKey defined on their own already. Maybe we could flip it on its head and say that a DirectAnswer type is a `FeaturedSnippetDirectAnswer | FieldValueDirectAnswer` and that would let us specify just DirectAnswer in UniversalSearchResponse? (feel like I could be totally off mark here)

cea2aj: The extends keyword is typically used so that interfaces can inherit properties from another interface without having to repeat them. That means we could remove the duplicated properties from the other direct answers if we wanted. I put them in there so that they show up in the docs (Nan and I decided on that pattern when making the autocomplete models). Because of that, we could remove the `DirectAnswer` model altogether, but I kind of like having the model to show the common properties)

I agree that it would be nice to have the UniversalSearchResponse return a DirectAnswer, however having the two models explicitly means that typescript forces you to check the type for accessing the properties specific to one of the models, which is really nice. If you don't check the type, Typescript lets you access the common properties. If we have a DirectAnswer which contains all of the possible properties, we would lose that functionality

oshi97 (approved):
lgtm! (it looks like tom and christian have already taken a look?)

creotutar (approved):
✂️ ✂️ ✂️ LGTM

PR #70, opened 2021-03-30
Handle highlighted fields with arrays

oshi97:
This commit adds support for highlighted fields that contain
array data. Previously, we would incorrectly use the array index as the fieldName.

J=SLAP-1194
TEST=auto,manual

hook up local SDK to my local core and see correct highlighted fields for
mt sinai, see highlighted fields being parsed correctly

### DIFF
--- a/src/transformers/searchservice/HighlightedValueFactory.ts
+++ b/src/transformers/searchservice/HighlightedValueFactory.ts
@@ -37,12 +37,17 @@ export class HighlightedValueFactory {
 
     const highlightedValues: HighlightedValue[] = [];
 
-    Object.entries(data).forEach(([fieldName, highlightedField]) => {
+    Object.entries(data).forEach(([fieldName, highlightedFieldData]) => {
       const currentPath = [...path];
       currentPath.push(fieldName);
 
-      if (this.isChildHighlightedField(highlightedField)) {
-        const { value, matchedSubstrings } = highlightedField;
+      if (Array.isArray(highlightedFieldData)) {
+        const highlightedValuesFromArray = highlightedFieldData.map(({ value, matchedSubstrings }) => {
+          return this.from(value, fieldName, currentPath, matchedSubstrings);
+        });
+        highlightedValues.push(...highlightedValuesFromArray);
+      } else if (this.isChildHighlightedField(highlightedFieldData)) {
+        const { value, matchedSubstrings } = highlightedFieldData;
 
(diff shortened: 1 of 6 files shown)


src/transformers/searchservice/HighlightedValueFactory.ts
```diff
@@ -34,24 +34,22 @@ export class HighlightedValueFactory {
     if (typeof data !== 'object' || data === null){
       return [];
     }
+    if (this.isChildHighlightedField(data)) {
+      return [ this.from(data.value, path[path.length - 1], path, data.matchedSubstrings) ];
```
cea2aj, line 38: nit: adding a `const fieldName = path[path.length - 1];` and passing that in as the second parameter might make this a little easier to understand

oshi97: yeah I like it! I was actually rereading my code and had to do a double take on why I did this lol

tests/transformers/HighlightedValueFactory.ts
```diff
@@ -86,4 +86,125 @@ it('Nested highlighted fields', () => {
...
+      ]
+    }
+  ];
+
+  const actualHighlightedValues = HighlightedValueFactory.create(liveApiData);
+  expect(actualHighlightedValues).toMatchObject(expectedHighlightedValues);
+});
+
+it('Highlighted fields with arbitrarily placed arrays', () => {
+  const liveApiData = {
```
cea2aj, line 135: Can you add a nested field inside one of array items?

src/transformers/searchservice/HighlightedValueFactory.ts
```diff
@@ -34,24 +34,23 @@ export class HighlightedValueFactory {
...
+    } else {
+      Object.entries(data).forEach(([fieldName, highlightedField]) => {
+        const currentPath = [...path, fieldName];
+        const nestedHighlightedValues = this.createRecursively(highlightedField, currentPath);
         highlightedValues.push(...nestedHighlightedValues);
-      }
-    });
-
+      });
+    }
```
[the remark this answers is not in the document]
oshi97: I wanted to use flatMap here but we would need to make tweaks to our tsconfig for es2019

tests/transformers/HighlightedValueFactory.ts
oshi97: yeah - nested arrays was technically being handled by the implementation, but I thought it wasn't possible so didn't add a test case for it. After talking with Nathan clearly it is possible!

PR #73, opened 2021-03-31
Replace HighlightedValues[] with HighlightedFields

oshi97:
This commit replaces the highlightedValues portion of Result
with highlightedFields. HighlightedFields is just a pass through
of the highlightedFields api response.

TEST=manual

hooked up local answers-core to SDK and then mt. sinai.
saw highlighted fields being parsed correctly


etc/answers-core.api.md
```diff
@@ -201,16 +201,19 @@ export interface FilterSearchResponse {
 }
 
 // @public
-export interface HighlightedValue {
-    fieldName: string;
+export interface HighlightedFieldLeaf {
```
cea2aj, line 204: I'm thinking that I may actually prefer keeping it named HighlightedValue. Do you remember if we discussed any reasons why we shouldn't do that??

oshi97: don't think we discussed what to name this one, I can update to HighlightedValue

src/models/searchservice/response/HighlightedFields.ts
```diff
@@ -0,0 +1,41 @@
...
+ * ```
+ *
+ * @public
+ */
+export interface HighlightedFields {
+  /**
+   * A mapping of a field to either an array of HighlightedFields, HighlightedFields for
+   * this field's subfields, or a {@link HighlightedFieldLeaf} for the field.
+   */
+  [fieldId: string]: HighlightedFields[] | HighlightedFields | HighlightedFieldLeaf
```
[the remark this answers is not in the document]
oshi97: ah yes I missed that

src/models/searchservice/response/HighlightedFieldLeaf.ts
```diff
@@ -1,30 +1,9 @@
...
-   * {
-   *   description: {
-   *     featured: 'The offical answers engine'
-   *   }
-   * }
-   * ```
-   * The associated path would then be `['description', 'featured']`
-   */
-  path: string[],
+ export interface HighlightedFieldLeaf {
```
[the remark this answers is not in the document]
oshi97: I can rename to HighlightedValue, does that work? Calling it a leaf made it a little more obvious to me what the difference between HighlightedValue and HighlightedFields were, but I agree with not exposing impl details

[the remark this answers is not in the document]
oshi97: Though, I guess in a way it's not really an implementation detail? Since the models are the interface

src/models/searchservice/response/HighlightedFields.ts
```diff
@@ -0,0 +1,41 @@
+import { HighlightedValue } from './HighlightedValue';
+
+/**
+ * A mapping of the values emphasized by the Answers API.
```
cea2aj, line 4: nit: I would maybe add something like: The structure of HighlightedFields matches the structure of the corresponding rawData of a {@link Result}

src/models/searchservice/response/HighlightedFields.ts
```diff
@@ -0,0 +1,41 @@
...
+ * ```
+ *
+ * @public
+ */
+export interface HighlightedFields {
+  /**
+   * A mapping of a field to either an array of HighlightedFields, HighlightedFields for
+   * this field's subfields, or a {@link HighlightedValue} for the field.
+   */
+  [fieldId: string]: HighlightedValue | HighlightedValue[] | HighlightedFields | HighlightedFields[]
```
cea2aj, line 40: It doesn't look like the api-documenter supports this data shape in the generated docs, so I think we want to explain the data type at the top under an `@remarks` tag. I would put something along the lines of: "HighlightedFields is an object of arbitrary shape which contains `{@link HighlightedValue}` objects.

oshi97: updated!

src/models/searchservice/response/Result.ts
```diff
@@ -35,8 +35,8 @@ export interface Result {
    * distanceFromFilter value will be from the search result to 'New York'.
    */
   distanceFromFilter?: number;
-  /** An array of {@link HighlightedValue | highlighted values} associated with the result. */
-  highlightedValues?: HighlightedValue[];
+  /** A tree of {@link HighlightedFields | highlighted fields} emphasized by the api. */
```
tmeyer2115, line 38: I think we should leave out the word "tree" here. At the top-level, the `highlightedFields` is an object, keyed by fieldId. The phrase "tree" may lead someone to believe that there's a single, root node in `highlightedFields`, when in fact there may be many top-level entries.

src/models/searchservice/response/HighlightedFields.ts
```diff
@@ -0,0 +1,47 @@
+import { HighlightedValue } from './HighlightedValue';
+
+/**
+ * A mapping of the values emphasized by the Answers API.
+ *
+ * @remarks
+ * The structure of HighlightedFields matches the structure of the corresponding rawData of a {@link Result}.
```
tmeyer2115, line 7: I think it's important to caveat this. The structure of `HighlightedFields` doesn't necessarily match the structure of `rawData`. Not every field that appears in `rawData` will have highlighting. The fields that do have highlighting will have the same structure in here that they do in `rawData`.

src/models/searchservice/response/HighlightedFields.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @example
+ * If a user searches for 'pet' and a description field in the results contains the value 'A pet store',
+ * the API will likely match the word 'pet'.
```
tmeyer2115, line 11: In the generated docs, the examples are a little confusing. Looking at the docs, I saw the JS snippet from the second example and immediately assumed it was related to the first example, which had no snippet. Then, I saw the two examples were unrelated. The first example doesn't have a sample JS snippet and the second doesn't have much introductory explanation.

src/models/searchservice/response/Result.ts
oshi97: Hmm I feel like this is still a tree, since the single root node is `highlightedFields`, but I think if somebody wants to know about the structure they can probably just click the link

src/models/searchservice/response/HighlightedFields.ts
oshi97: agree!

src/models/searchservice/response/HighlightedFields.ts
oshi97: ah yeah I can unify the two

src/models/searchservice/response/HighlightedFields.ts
cea2aj: It looks like the docs generation does actually support this if we define it as a type rather than an interface. Can you try that? I think it would look like `type HighlightedFields = { [fieldId: string] : HighlightedValue | HighlightedValue[] | HighlightedFields | HighlightedFields[] }`

src/models/searchservice/response/HighlightedFields.ts
```diff
@@ -0,0 +1,49 @@
+import { HighlightedValue } from './HighlightedValue';
+
+/**
+ * A mapping of the values emphasized by the Answers API.
+ *
+ * @remarks
+ * Only fields that have been marked as highlighted will be present - which may not be
+ * all fields of the corresponding {@link Result}'s rawData.
+ * The fields the are present will match structure and order of the rawData.
```
cea2aj, line 9: "fields that are present"

oshi97: oof

src/models/searchservice/response/HighlightedFields.ts
oshi97: let me try

src/models/searchservice/response/HighlightedFields.ts
```diff
@@ -4,30 +4,32 @@ import { HighlightedValue } from './HighlightedValue';
...
  * @remarks
- * The structure of HighlightedFields matches the structure of the corresponding rawData of a {@link Result}.
+ * Only fields that have been marked as highlighted will be present - which may not be
+ * all fields of the corresponding {@link Result}'s rawData.
+ * The fields the are present will match structure and order of the rawData.
  *
  * @example
- * If a user searches for 'pet' and a description field in the results contains the value 'A pet store',
- * the API will likely match the word 'pet'.
+ * If a user searches for 'apple' and a description field in the results contains the value 'apple pie',
```
tmeyer2115: Update the description here to match the JS snippet. The snippet has fields` name` and `c_favoriteFruits`, but no `description` field.

src/models/searchservice/response/HighlightedFields.ts
oshi97: works!

src/models/searchservice/response/HighlightedFields.ts
oshi97: already updated! sorry stumbled on my words

src/models/searchservice/response/HighlightedFields.ts
cea2aj: Now it looks like the description and remarks within the type doesn't show up in the docs. I'm not sure that we need the section "A mapping of a field to either..." anymore since the shape is now described in the docs. The remark "HighlightedFields is an object of arbitrary shape..." may fit at the top of the current remarks. Thoughts?

src/models/searchservice/response/HighlightedFields.ts
```diff
@@ -0,0 +1,58 @@
+import { HighlightedValue } from './HighlightedValue';
+
+/**
+ * A mapping of the values emphasized by the Answers API.
```
cea2aj, line 4: nit: What do you think about updating this to "A mapping of fields to the values emphasized by the Answers API"?

oshi97: yeah I like that

tmeyer2115 (approved):

PR #74, opened 2021-04-05
Document the library builds

cea2aj:
Add some documentation which give a better overview of the
different builds, and how to use them

J=SLAP-1174
TEST=visual

Make sure that the README looks as expected

PR #75, opened 2021-04-12
v1.1.0

tmeyer2115:
### Changes
- Added documentation for the different builds of the library that we maintain. These descriptions should help clients make a more informed choice about which build best fits their use case. (#74)
- Added support for `FEATURED_SNIPPET` type direct answers. (#69)

### Bug Fixes
- The `queryId` attribute was added to the `VerticalSearchRequest` model. This attribute is necessary for pagination to work properly. (#68)
- Fixed a bug that caused `Result`s from non-KMG backends to be formatted improperly. (#72)
- The `highlightedValues` attribute of a `Result` has been removed. In its place is a new attribute, called `highlightedFields`, that is of type `HighlightedFields`. The old attribute was not adequate to represent the complexity of the API response. (#73)


cea2aj (approved):
lgtm

PR #78, opened 2021-05-06
add Coveralls code coverage

oshi97:
Adds a Coveralls github workflow, also adds a coverage badge
to the readme. I tried looking online for a way to make the coverage
badge reflect the current branch, instead of always master,
but it sounds like there's no way to do that with a static readme.

J=SLAP-873
TEST=auto

we have a coveralls check now (see below) that displays the code coverage percent, wow!
clicking into it links to the coverage for the build, double wow!

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,4 +1,3 @@
-
 # This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -4,7 +4,10 @@
 
 name: Code Coverage
 
-on: [push, pull_request]
+on:
+  push:
+    branches-ignore: dev/*
+  pull_request:
 
 jobs:
   coverage:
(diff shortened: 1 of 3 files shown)


cea2aj (approved):
🚀  triple wow!

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
```
tmeyer2115, line 1: Could we mention somewhere in the comment that code coverage info will be included?

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,30 @@
...
+        node-version: [16.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npx jest --coverage
```
tmeyer2115, line 26: For our PRs, will this cause us to be running through the unit tests twice?

oshi97: yeah it will run our unit tests one extra time. Currently they are run for 6 different permutations, with 3 different node versions, and the git branch being either the pushed branch, or the result of the pushed branch being merged with the base branch. I can look into trying to combine this with our run-tests.yml

.github/workflows/coverage.yml
oshi97: yes, updated!

PR #79, opened 2021-05-12
add AnswersConfig.additionalQueryParams? property

oshi97:
This commit adds an additionalQueryParams prop to AnswersConfig.
These additional params will be passed as
query params to every single request made by answers-core.

Because this property is marked as alpha, it is not exported as part of the
AnswersConfig interface, and any typescript users of core will receive
a compilation error if they try to use this property. This property can
still be used in regular js usages of core. In a following PR, we will
add an untrimmed bundle.d.ts which will include alpha and internal
types.

J=SLAP-1320
T=405448
TEST=manual,auto

use local core in the SDK, see that jsLibVersion is passed in search, autocomplete,
and qa submission requests
also try in answers-core-testing

### DIFF
--- a/tests/infra/AutoCompleteServiceImpl.ts
+++ b/tests/infra/AutoCompleteServiceImpl.ts
@@ -129,21 +129,16 @@ describe('additionalQueryParams are passed through', () => {
       jsLibVersion: 'LIB_VERSION'
     }
   };
-  const mockHttpService = new HttpServiceMock();
-  const apiResponseValidator = new ApiResponseValidator();
+  let mockHttpService, apiResponseValidator;
+  beforeEach(() => {
+    mockHttpService = new HttpServiceMock();
+    apiResponseValidator = new ApiResponseValidator();
+  });
 
   it('Universal Autocomplete', async () => {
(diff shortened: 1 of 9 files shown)


tests/infra/AutoCompleteServiceImpl.ts
```diff
@@ -119,4 +118,102 @@ describe('AutocompleteService', () => {
...
+  };
+  const mockHttpService = new HttpServiceMock();
+  const apiResponseValidator = new ApiResponseValidator();
+
+  it('Universal Autocomplete', async () => {
+    mockHttpService.get.mockResolvedValue(mockAutocompleteResponse);
+    const request: UniversalAutocompleteRequest = {
+      input: ''
+    };
+    const expectedQueryParams = {
```
cea2aj, line 140: What are your thoughts on using a jest expect.objectContaining for these tests which just checks for the jsLibVersion?

oshi97: yeah I like that more! no point to be redundant

tmeyer2115 (approved):

PR #80, opened 2021-05-13
add internal bundle

oshi97:
The internal bundle lets TS users use internal
and alpha types. internal and alpha types are
still exported with the lib/esm and lib/commonjs
bundles, SLAP-1325 was made to address those
at a later date.

J=SLAP-1321
TEST=manual

test that I can use the internal bundle in the
core testing lib

### DIFF
--- a/internal/package.json
+++ b/internal/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "answers-core-legacy",
+  "name": "answers-core-internal",
   "main": "../lib/esm/index.js",
   "types": "../dist/bundle-internal.d.ts"
 }
--- a/internal/package.json
+++ b/internal/package.json
@@ -1,5 +1,6 @@
 {
   "name": "answers-core-internal",
-  "main": "../lib/esm/index.js",
+  "main": "./lib/commonjs/index.js",
+  "module": "./lib/esm/index.js",
   "types": "../dist/bundle-internal.d.ts"
 }
--- a/internal/package.json
+++ b/internal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "answers-core-internal",
-  "main": "./lib/commonjs/index.js",
-  "module": "./lib/esm/index.js",
+  "main": "../lib/commonjs/index.js",
+  "module": "../lib/esm/index.js",
   "types": "../dist/bundle-internal.d.ts"
 }
--- a/internal/package.json
+++ b/internal/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "answers-core-legacy",
+  "main": "../lib/esm/index.js",
+  "types": "../dist/bundle-internal.d.ts"
+}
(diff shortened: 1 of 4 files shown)


internal/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "answers-core-legacy",
```
cea2aj, line 2: Should this be answers-core-internal?

oshi97: yes, updated!

cea2aj:
I thought we were simply going to publish the internal version of core with an alpha tag to NPM. What was your reasoning behind adding an internal bundle?

oshi97:
> I thought we were simply going to publish the internal version of core with an alpha tag to NPM. What was your reasoning behind adding an internal bundle?

We are publishing an alpha tag to NPM. That alpha tag will contain an additional internal bundle.d.ts
The reasoning is that, if we have an internal bundle, we can reference that internal bundle in any TS app we want, forever. 
If we update the main bundle.d.ts for just this alpha tag, it cannot be merged into develop. If we need to add more internal only features in the future, we will need to make this kind of change again.

cea2aj:
> > I thought we were simply going to publish the internal version of core with an alpha tag to NPM. What was your reasoning behind adding an internal bundle?
> 
> We are publishing an alpha tag to NPM. That alpha tag will contain an additional internal bundle.d.ts
> The reasoning is that, if we have an internal bundle, we can reference that internal bundle in any TS app we want, forever.
> If we update the main bundle.d.ts for just this alpha tag, it cannot be merged into develop. If we need to add more internal only features in the future, we will need to make this kind of change again.

My understanding of the suggested release flow by api-extractor for trimmed versions is that we update the bundled type files and publish that, however that change isn't supposed to get committed into the repo. The documentation mentions that some people create scripts which automate this. We could commit such a script to the repo.

The downside of adding a new bundle type for internal is that the the legacy bundle won't include the internal types, even when using the alpha tag from NPM. Also Product wants to keep the number of bundles we maintain to a minimum, and we should check with them on this.

oshi97:
> > > I thought we were simply going to publish the internal version of core with an alpha tag to NPM. What was your reasoning behind adding an internal bundle?
> > 
> > 
> > We are publishing an alpha tag to NPM. That alpha tag will contain an additional internal bundle.d.ts
> > The reasoning is that, if we have an internal bundle, we can reference that internal bundle in any TS app we want, forever.
> > If we update the main bundle.d.ts for just this alpha tag, it cannot be merged into develop. If we need to add more internal only features in the future, we will need to make this kind of change again.
> 
> My understanding of the suggested release flow by api-extractor for trimmed versions is that we update the bundled type files and publish that, however that change isn't supposed to get committed into the repo. The documentation mentions that some people create scripts which automate this. We could commit such a script to the repo.
> 
> The downside of adding a new bundle type for internal is that the the legacy bundle won't include the internal types, even when using the alpha tag from NPM. Also Product wants to keep the number of bundles we maintain to a minimum, and we should check with them on this.

Those are good points. We still want some long term solution for an internal bundle but can shelve for now.

PR #81, opened 2021-05-13
make Endpoints public

oshi97:
This commit exposes the Endpoints type and
AnswersConfig.endpoints? property to the
public bundle.

J=SLAP-1323
TEST=manual

see that I can use endpoints in the core testing lib


src/models/core/Endpoints.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * URLs to various Answers services.
```
cea2aj, line 2: nit: We may want to provide some more info now that this is public. Maybe something like: "Overrides for the URLs which are used when making requests to the Answers API "

cea2aj (approved):
looks good!

PR #84, opened 2021-06-14
Add 'fieldType' to the DirectAnswer model

cea2aj:
Add a `fieldType` property to the `DirectAnswer` model

This allows us to check for the `fieldType` of 'rich_text' for `FeaturedSnippetDirectAnswer`s
 
J=SLAP-1383
TEST=manual, auto

Update test coverage to include the `fieldType` property. Also use the test site to confirm that the `fieldType` property is appearing for `FeaturedSnippetDirectAnswer`s


src/models/searchservice/response/FeaturedSnippetDirectAnswer.ts
```diff
@@ -17,6 +17,8 @@ export interface FeaturedSnippetDirectAnswer extends DirectAnswer {
   relatedResult: Result;
```
tmeyer2115, line 17: Do we need to mark the `value` above as optional? Or does that come from the `inheritDoc`?

tmeyer2115 (approved):

PR #94, opened 2021-07-01
Added session id to core

cjiang2000:
J=SLAP-1408
TEST=manual

Checked on test page to make sure that session_id was being passed properly.


src/models/searchservice/request/UniversalSearchRequest.ts
```diff
@@ -16,6 +16,8 @@ export interface UniversalSearchRequest {
   queryTrigger?: QueryTrigger;
   /** Disables spellcheck if true. */
   skipSpellCheck?: boolean;
+  /** The session ID*/
```
cea2aj, line 19: nit: missing a space after 'ID'

cea2aj: also, for the documentation in this repo, we always put a period at the end because it works better in the generated docs

src/models/searchservice/request/VerticalSearchRequest.ts
```diff
@@ -33,6 +33,8 @@ export interface VerticalSearchRequest {
   location?: LatLong,
   /** {@inheritDoc QueryTrigger} */
   queryTrigger?: QueryTrigger,
+  /** The session ID */
+  sessionId?: string,
```
cea2aj, line 37: Can we add a little more to this description? The Yext API docs describe this as: 'UUID used to track session state when cookies are blocked.' Maybe we can write something similar such as 'Used to track session state when cookies are blocked.'?

oshi97 (approved):

PR #95, opened 2021-07-12
Allow arbitrary string query sources

cea2aj:
Allow an arbitrary string to be supplied as a query source for vertical and universal searches

This allows us to supply custom sources such as `SUPPORT_SEARCH`

J=SLAP-1438
TEST=manual, auto

Add a unit test for the functionality. Also use the core test site to supply arbitrary strings and observe it being sent in network requests

### DIFF
--- a/src/infra/SearchServiceImpl.ts
+++ b/src/infra/SearchServiceImpl.ts
@@ -34,7 +34,7 @@ interface UniversalSearchQueryParams extends QueryParams {
   queryTrigger?: QueryTrigger,
   context?: string;
   referrerPageUrl?: string,
-  source?: QuerySource
+  source?: QuerySource | string
 }
 
 /**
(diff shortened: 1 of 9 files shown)


oshi97 (approved):

src/infra/SearchServiceImpl.ts
```diff
@@ -34,7 +34,7 @@ interface UniversalSearchQueryParams extends QueryParams {
   queryTrigger?: QueryTrigger,
   context?: string;
   referrerPageUrl?: string,
-  source?: QuerySource
+  source?: QuerySource | string
```
tmeyer2115, line 37: A minor thing, but could we add `SUPPORT_SEARCH` to the `QuerySource` enum? I know it's somewhat irrelevant now that any old string will work. But, I'd like `QuerySource` to list all the "official" or "out of the box" sources so to speak.

cea2aj: I asked Rose about this and she felt that 'SUPPORT_SEARCH' is a custom solution thing, so she didn't want it added to the enum

tmeyer2115 (approved):

PR #97, opened 2021-07-29
pass credentials: 'include' in get requests and simplify HttpService

oshi97:
GET requests were missing the credentials: 'include' header, which was
causing cookies to not be sent.

This commit
- changes HttpServiceImpl to no longer let you pass in specific overrides for RequestInit.
- Tries to use window.fetch if possible, which is what the SDK used to do. Before, we would always use cross-fetch, which will always use its whatwg-fetch ponyfill. It looks like there is some issue where using whatwg-fetch will prevent cookies from being displayed properly in the chrome/edge dev tools "Application -> Cookies" tab. 
   - Even with this change I can't get cookies to show up in Firefox/Safari dev tools, though they are still sent properly in the request. I also cannot get cookies to show up with SDK v1.7 on those browsers' dev tools so might be a personal browser settings thing.
- removes jest-fetch-mock because it was not needed after we added cross-fetch

Originally I intended for usages of HttpServiceImpl to manually pass
in request headers, and to not provide any defaults outside of the bare
minimum. Looking back, this added unneeded complexity to HttpService,
since all GET requests in 1.7 of answers-search-ui used the same headers,
and we only had one possible POST request.

T=TECHOPS-1384
TEST=manual

hooked up my local answers-core to answers-search-ui
saw that cookies were being sent in search requests, and that
question submission still had the same cors and content-type headers

tested on safari, chrome, firefox, edge, ie11

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -59,7 +59,7 @@ THE SOFTWARE.
 
 The following NPM packages may be included in this product:
 
- - cross-fetch@3.0.6
+ - cross-fetch@3.1.4
 
 These packages each contain the following license and notice below:
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -5244,16 +5244,6 @@
         "jest-util": "^26.6.2"
       }
     },
-    "jest-fetch-mock": {
-      "version": "3.0.3",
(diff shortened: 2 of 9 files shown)


cea2aj (approved):
looks good. Do you think it might also be worth checking IE11?

oshi97:
> looks good. Do you think it might also be worth checking IE11?

Good point, checked and its working!

PR #90, opened 2021-06-25
Added limit for universal search requests.

cjiang2000:
J=SLAP-1267
TEST=manual/auto

Checked answers-core-test site and sdk for new limit in action.

PR #99, opened 2021-08-24
Version 1.3

cea2aj:
### Features
- Direct answers are now supported on vertical searches (#96)
- For universal searches, a limit for each vertical can now be supplied (#90) (#93)
- A session ID can now be supplied for vertical and universal searches (#94)
- A query source can now be an arbitrary string for vertical and universal searches (#95)
### Enhancements
- Question submissions are now sent to a new endpoint with higher availability (#98)
### Bugfixes
- Fixes a bug which caused cookies to not be sent properly (#97)


yen-tt:
package version update?

yen-tt (approved):

cea2aj:
> package version update? or I guess that can be done when pushing the tag later. lgtm!

Good catch, i'll add it now

oshi97 (approved):

PR #100, opened 2021-08-24
Bump version to 1.3.0

cea2aj:


oshi97 (approved):
think we generally don't need PRs for these anymore, but lgtm!

PR #103, opened 2021-09-15
Fix node compatibility issue

cea2aj:
Fix the window access issue which was causing the library to break on node

I added a unit test which ensures that the HttpService works on node

J=none
TEST=manual, auto

Test on the browser and on node and confirm searches work on both

### DIFF
--- a/src/infra/HttpServiceImpl.ts
+++ b/src/infra/HttpServiceImpl.ts
@@ -58,7 +58,7 @@ export class HttpServiceImpl implements HttpService {
     reqInit: RequestInit
   ): Promise<Response> {
     const urlWithParams = addParamsToURL(url, queryParams);
-    if (window.fetch) {
+    if (typeof(window) !== 'undefined' && window.fetch) {
       return window.fetch(urlWithParams, reqInit);
     }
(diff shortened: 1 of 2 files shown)


tmeyer2115 (approved):

src/infra/HttpServiceImpl.ts
```diff
@@ -58,7 +58,7 @@ export class HttpServiceImpl implements HttpService {
     reqInit: RequestInit
   ): Promise<Response> {
     const urlWithParams = addParamsToURL(url, queryParams);
-    if (window.fetch) {
+    if (typeof(window) !== 'undefined' && window.fetch) {
```
oshi97, line 61: would globalThis.fetch work instead of window.fetch?

cea2aj: globalThis doesn't have IE support. We could polyfill it, but that seems a little overkill. Thoughts?

oshi97: ah ok was just wondering if it would be slightly nicer

oshi97 (approved):

PR #109, opened 2021-10-05
Visitor Support

cea2aj:
Add the visitor option to the AnswersConfig which will be included with every request with the exception of question submission requests


src/infra/AutocompleteServiceImpl.ts
```diff
@@ -70,6 +70,7 @@ export class AutocompleteServiceImpl implements AutocompleteService {
       version: this.config.experienceVersion,
       locale: this.config.locale,
       sessionTrackingEnabled: request.sessionTrackingEnabled,
+      visitor: JSON.stringify(this.config.visitor || undefined),
```
yen-tt, line 73: if this.config.visitor is not defined, it would automatically be JSON.stringify(undefined) so I think we can remove `|| undefined`

src/models/core/Visitor.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @public
+ */
+export interface Visitor {
+  /** The ID associated with the user */
+  id: string,
+  /** The type of visitor
+   *
+   * @example 'YEXT_USER' for Yext Auth
+   */
+  idMethod: string
```
yen-tt, line 16: Do we expect idMethod to always be provided with visitor data? I think spec mark this as optional

cea2aj: yep, it should be optional. Updating

PR #111, opened 2021-10-08
Use api-extractor's type bundle for all builds

cea2aj:
Use api-extractor's bundle.d.ts for all builds of the library

We want to use api-extratcor's bundle.d.ts so that fields marked as alpha, beta, or internal won't be present in output types. Accomplish this by copying the type bundle from dist/bundle.d.ts into lib/esm and lib/commonjs.

Since api-extractor does not yet support bundled declaration maps, I disabled declarationMap in tsconfig.base.json. If we were to leave it on, users of the core library would be incorrectly taken to index.js whenever they click 'Go to Definition' on one of our types or interfaces. This is a confusing developer experience because index.js does not contain the corresponding type definition. I found that if we disable declarationMap, developers will instead be taken to the correct interface inside of index.d.ts.

J=SLAP-1325
TEST=manual

Use the core test site and confirm that a field marked as alpha (additionalQueryParams from AnswersConfig) does not appear in the autocomplete options when using the AnswersConfig model. Also test clicking the "Go to Definiton" button and see it take me to the corresponding model in index.d.ts

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,7 @@
         "ts-loader": "^8.0.14",
         "typescript": "^4.0.3",
         "webpack": "^5.18.0",
-        "webpack-cli": "^4.9.0"
+        "webpack-cli": "^4.4.0"
       },
       "engines": {
         "node": ">=0.12"
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "ts-loader": "^8.0.14",
     "typescript": "^4.0.3",
     "webpack": "^5.18.0",
-    "webpack-cli": "^4.9.0"
+    "webpack-cli": "^4.4.0"
   },
   "dependencies": {
     "@babel/runtime-corejs3": "^7.12.5",


package-lock.json
```diff
@@ -1,12056 +1,8 @@
 {
   "name": "@yext/answers-core",
   "version": "1.3.2",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
nmanu1, line 4: I think we want to keep lockfileVersion 2

package.json
```diff
@@ -20,7 +20,7 @@
     "tsc-esm": "tsc -p tsconfig.esm.json",
     "tsc": "npm run tsc-cjs && npm run tsc-esm",
     "build-legacy-bundle": "npx webpack",
-    "build": "npm run tsc && npm run build-legacy-bundle && npm run api-extractor && npm run generate-docs",
+    "build": "npm run tsc && npm run build-legacy-bundle && npm run api-extractor && ./scripts/setup-lib-types.sh &&npm run generate-docs",
```
oshi97, line 23: nit: space after &&

package-lock.json
cea2aj: Node 8 just came out and I think we are good to use that one as well since it's very similar to 7 except it drops support for some of the older node versions. I'm not sure how we'd setup npm 7 and node 14, however we can easily setup npm 8 and node 14 with: `nvm use 14 && nvm install-latest-npm`

PR #112, opened 2021-11-10
Update FilterSearchResponse

yen-tt:
- update `FilterSearchResponse` to remove top level fields: `sectioned`, `results` and `inputIntents` as those are not returned by the backend. All results, sectioned or not, will be in `sections` field. (from [liveapi doc](https://hitchhikers.yext.com/docs/liveapis/answersapi/verticalsearch/#operation/filtersearch))
- when `failedVerticals` is not empty in `FilterSearchResponse`, throw AnswersError
- update tests 

Note: this will require update in SDK

J=SLAP-1689
TEST=auto

see that jest test passed


src/models/autocompleteservice/AutocompleteResponse.ts
```diff
@@ -24,29 +25,41 @@ export interface AutocompleteResponse {
...
-   * An array of {@link AutocompleteResult}s.
-   *
-   * @remarks
-   * This array will only be populated if sectioned is false.
-   */
-  results: AutocompleteResult[],
-  /** {@inheritDoc SearchIntent} */
-  inputIntents: SearchIntent[];
+  /** Contains error information when one or more verticals fail to return results. */
+  failedVerticals: {
```
tmeyer2115, line 46: @mgdunn2 do you think this is something worth exposing to a user of Core? The concept of a `failedVertical` for filter search seems kind of funny. It makes sense for something like universal search. Is this deprecated?

mgdunn2: At the moment if there is an error in filter search it will be returned as a failed vertical. Since filter search is now being scoped to only a single vertical it might make sense to just fail the request entirely rather than return a 200 with a failed vertical but this does reflect the current state of the API.

tmeyer2115 (approved):

PR #114, opened 2021-11-15
Add AnswersConfigWithToken

cea2aj:
~~This pr aims to solve a breaking change of using types to defined config to require either token or apikey, since types can't be extended. Instead, this pr will keep the original AnswersConfig that uses apiKey only and add an additional config `AnswersConfigWithToken` to supports config with a token rather than an api key~~
After meeting with product, it's decided that breaking change of using types is fine to use. This pr add some tweaks to the config: removing the optional fields in baseAnswersConfig, and some type guard in the services

J=SLAP-1680
TEST=auto

see that it compiled successfully and the updated jest tests passed
have SDK use local core changes, see that request have apiKey in request query string parameters when only apiKey is supplied. see that request have token in authorization header when only token is supplied


oshi97 (approved):

src/models/core/AnswersConfig.ts
```diff
@@ -42,44 +40,12 @@ export interface BaseAnswersConfig {
...
-export interface AnswersConfigWithToken extends BaseAnswersConfig {
-  /**
-   * {@inheritDoc BaseAnswersConfig.token}
-   */
-  token: string,
-  /**
-   * apiKey should NOT be provided along with token
-   */
-  apiKey?: never
+export interface AnswersConfigWithToken extends Omit<AnswersConfig,'apiKey'> {
```
oshi97, line 48: nit: space after comma

tmeyer2115: I'm not a big fan of having an `AnswersConfig` and an `AnswersConfigWithToken`. As a consumer, I'd expect there to be a single configuration object. It feels a bit odd to make a whole new configuration object that's exactly the same, except with one property changed.

tmeyer2115 (commented):
I'm not a big fan of having an `AnswersConfig` and an `AnswersConfigWithToken`. As a consumer, I'd expect there to be a single configuration object. It feels a bit odd to make a whole new configuration object that's exactly the same, except with one property changed.

src/models/core/AnswersConfig.ts
oshi97: maybe could do some type composition like the below?
```ts
// unsure about all naming
interface AnswersApiKey = { apiKey: string }
interface AnswersToken = { token: string }
export type AnswersConfighWithAuth = (AnswersConfig & AnswersApiKey) | (AnswersConfig & AnswersToken)
```

yen-tt: there's two other ways we could do:
1) have both in AnswersConfig as optional. But this loses the typescript check and we would have to manually check and throw error to user in the different services if both are defined or undefined.
```
 apiKey?: string
 token?: string
```
2) have something similar to visitor. Where both fields are required, and user specified what to use. But this wouldn't be backward compatible.
```
authMethod: 'API_KEY' or 'TOKEN'
authId: string
```

yen-tt: @oshi97, that would uses type instead of interface, which is the breaking change that connor was talking about. I'm not sure if we would want that?

oshi97: I'm okay with a breaking change. This isn't the best justification, but I believe that few people are using this library besides us, and imagine that 100% of them are within Yext. I agree with Tom that `AnswersConfigWithToken` is a little strange.
In my mind the goals here are
1. refactor `AnswersConfig` in a way that won't need to be broken later on (or at least broken less)
2. still have a nice interface for people to use. i.e., if we didn't have to worry about breaking anything, what would we want things to look like? 

To that end I think what we had before does that a little better, though I imagine there might be some improvements we can still make? I tried looking at some other libraries like the [firebase js sdk](https://github.com/firebase/firebase-js-sdk/blob/master/packages/app/src/public-types.ts) and [algolia's js client](https://github.com/algolia/algoliasearch-client-javascript/tree/master/packages/client-common/src/types) but couldn't really find anything obvious we were missing. Not sure if @cea2aj has any ideas here?

cea2aj: With the previous approach, we had 4 documentation pages for Answers configs (AnswersConfig, AnswersConfigWithToken, AnswersConfigWithApiKey, and BaseAnswersConfig). I figured this would be simpler with just 2 documentaiton pagees. Plus this allows us to remove the `never` keyword from the interfaces. The previous appraoch would enforce that only one type of auth is used for consumers of Core, however it would not do so within the library which is a bit strange. Here we have much stronger type checking at build time. We are also avoiding all breaking changes with this approach.

If we aren't concerned about breaking changes, then perhaps we could use the previous approach but with some modifications as you suggested. Maybe instead of using the 'never' keyword, we could use the `Omit` utility type to omit the auth method which isn't being used. That might give us the stronger typing as well

cea2aj: One idea I have for making the previous approach less of a breaking change would be to keep the `AnswersConfig` as an interface (but with apiKey and token marked as optional), and then we introduce a new type such as `SpecificAnswersConfig` which is the new `type` passed to `provideCore`

src/index.ts
```diff
@@ -17,10 +17,6 @@ import { ApiResponseValidator } from './validation/ApiResponseValidator';
  * @public
  */
 export function provideCore(config: AnswersConfig): AnswersCore {
-  if (config.apiKey && config.token) {
```
cea2aj, line 20: Can we do something like `if ('apiKey' in config && 'token' in config) {
    throw new Error('Both apiKey and token are present. Only one authentication method should be provided');
  }` so that we can keep this error message?

src/models/core/AnswersConfig.ts
```diff
@@ -65,9 +58,7 @@ export interface AnswersConfigWithApiKey extends BaseAnswersConfig {
  * @public
  */
 export interface AnswersConfigWithToken extends BaseAnswersConfig {
-  /**
-   * {@inheritDoc BaseAnswersConfig.token}
-   */
+  /** The authentication token of the answers experience. */
```
cea2aj, line 61: Maybe we should add to this so that it reads: "The authentication token of the answers experience which will be passed in the Auth header as a Bearer token"

src/models/core/AnswersConfig.ts
```diff
@@ -44,7 +44,7 @@ export interface BaseAnswersConfig {
  * @public
  */
 export interface AnswersConfigWithApiKey extends BaseAnswersConfig {
-  /** The api key of the answers experience. */
+  /** The api key of the answers experience which will be sent as a query param */
```
cea2aj, line 47: small nit: In core we always put a period at the end of the sentences because it works better with the generated docs

cea2aj:
Since I created this PR, I can't approve it, but LGTM!

PR #115, opened 2021-11-15
Change how the `Visitor` params are included in API requests.

tmeyer2115:
Originally, the spec called for the `Visitor` object to be passed as a single
param in the Live API requests. Now, the spec has been updated so that each
attribute of the `Visitor` is passed as its own top-level API param.

J=SLAP-1686
TEST=auto

Updated the unit tests and ensured they passed.


oshi97 (approved):

yen-tt:
Tested manually through SDK: added visitor object to ANSWERS.init and see that the visitorId and visitorIdMethod are two separate top-level fields in the request query string parameters when universal/vertical search and autocomplete is perform

oshi97 (approved):

PR #119, opened 2021-11-18
version bump

yen-tt:
bump version to publish a beta that include filterSearchResponse update

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yext/answers-core",
-  "version": "1.4.0",
+  "version": "1.4.0-beta.0",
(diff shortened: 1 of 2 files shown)


nmanu1 (approved):

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@yext/answers-core",
-  "version": "1.4.0",
+  "version": "1.4.0-beta.0",
```
cea2aj, line 3: Would it be more accurate to have 1.5.0-beta.0, since 1.4.0 is already released?

yen-tt: right

cea2aj (approved):

PR #122, opened 2021-11-23
add verticalKeys to AutocompleteResult

oshi97:
The work has not been done in the backend yet, so only automated testing could be done.

J=SLAP-1708
TEST=auto


tests/infra/AutocompleteServiceImpl.ts
```diff
@@ -33,6 +34,18 @@ describe('AutocompleteService', () => {
 
   const mockHttpService = new HttpServiceMock();
   const apiResponseValidator = new ApiResponseValidator();
+  function createMockAutocompleteService(params?: { response?: unknown, answersConfig?: AnswersConfig }) {
```
yen-tt, line 37: can the type be some union of the vertical and universal autocomplete response rather than unknown?

oshi97: oh hmm this is the raw api response so I don't think so. I considered making it
```ts
mockAutocompleteResponse | mockAutocompleteResponseWithSections | //etc
```
but didn't really see much value in that

yen-tt: oh I see, so can it be of type ApiResponse?

oshi97: ah yeah that's better!

yen-tt (approved):

PR #128, opened 2022-01-19
Add query rules actions

nmanu1:
Add data from new query rules actions to universal and vertical search responses.

J=SLAP-1799
TEST=auto

Check that the added jest tests for transforming the query rules response data pass.


yen-tt:
I believe you would need to add this new field as part of the return object in function `createVerticalSearchResponse` and `createUniversalSearchResponse`. Otherwise, you wouldn't see it in the response from `universalSearch` and `verticalSearch`. Probably worth updating the tests for searchServiceImpl to include this new field as well.

nmanu1:
> I believe you would need to add this new field as part of the return object in function `createVerticalSearchResponse` and `createUniversalSearchResponse`. Otherwise, you wouldn't see it in the response from `universalSearch` and `verticalSearch`. Updating the tests for searchServiceImpl to include this new field would be helpful to catch this.

Good catch! Updated those functions and their tests to include the new query rules field.

src/models/searchservice/response/QueryRulesActionsData.ts
```diff
@@ -0,0 +1,19 @@
...
+   */
+  key: string,
+  /**
+   * The data returned from the query rule.
+   */
+  data?: Record<string, unknown>,
+  /**
+   * Any errors returned from the query rule.
+   */
+  errors?: string
```
cea2aj, line 18: Do we know if this can return multiple errors? In general Rose prefers us to stick with the same name as the API which would mean changing this to a singular 'error'

src/models/searchservice/response/QueryRulesActionsData.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Data returned from the Answers query rules system.
+ *
+ * @public
+ */
+export interface QueryRulesActionsData {
+  /**
+   * The unique identifier for this query rule.
+   */
+  key: string,
```
oshi97, line 10: can this be truly any string? it would be nice if product had a list of the exact values we could get here, if possible. a little more structure would be nice for data too

tmeyer2115: Yes, this can be any string. The unique identifier for the query rule is decided by the customer, so it's arbitrary.

tmeyer2115: As for `data`, this is also dictated by the customer. I believe they can specify arbitrary JSON to be returned. It's a bit like the `rawData` case we already have in Core. It's on them to really know the structure of `data`.

src/models/searchservice/response/QueryRulesActionsData.ts
nmanu1: Yes, there can be multiple errors. I've updated the error object type after getting more clarification from the backend on what will be returned.

tmeyer2115 (approved):

PR #130, opened 2022-02-17
test github action

yen-tt:


.github/workflows/update_docs.yml
```diff
@@ -0,0 +1,27 @@
...
+        node-version: [17.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run build-ci
```
cea2aj, line 20: The build-ci script uses API extractor in production mode which doesn't update the API report file. I believe we should use the regular `npm run build` instead and consider removing build-ci

cea2aj: Another solution would be to update the messasge reporting config for API extractor to update the logLevel for the various message reporting types to 'error'. The main downside of this approach is that we wouldn't be able to build the docs locally unless we address all of the diagnostic messages

yen-tt: Alternatively, we could scan for warning/error messages and manually throw a non-zero exit in the github action. But it may be best to address the errors locally instead of waiting till github action fail. Updated using the second solution!

api-extractor.json
```diff
@@ -299,7 +299,7 @@
          *
          * DEFAULT VALUE: "warning"
          */
-        "logLevel": "warning"
+        "logLevel": "error"
```
cea2aj, line 302: I believe we also need to update `extractorMessageReporting` and `tsDocMessageReporting`

cea2aj (approved):
🚀

cea2aj (approved):

PR #131, opened 2022-03-02
Expose the various Service interfaces.

tmeyer2115:
This PR exposes the `SearchService`, `AutocompleteService`, and `QuestionSubmissionService` interfaces. It also makes the `AnswersCore` constructor public.

These changes will allow someone to create an `AnswersCore` instance directly, without having to use `provideCore`. It's possible that someone will want their own implementation of one of these services. This change will also help us better unit test our Component library.

J=SLAP-1928
TEST=none


src/AnswersCore.ts
```diff
@@ -18,7 +18,6 @@ import { AutocompleteService } from './services/AutocompleteService';
  * @public
  */
 export class AnswersCore {
-  /** @internal */
```
oshi97, line 21: for my edification was it already possible to instantiate your own AnswersCore, since we were already exporting the class, or did this annotation have say api-extractor prevent people from doing so via typescript yelling at them?

oshi97 (approved):

src/AnswersCore.ts
tmeyer2115: The constructor was exported but it took no arguments. I think it exported the "default"/"no params" constructor if that makes sense?

PR #141, opened 2022-03-14
Answers-Agent Support

nmanu1:
When hitting any endpoint, pass the Client-SDK header, which includes the ANSWERS_CORE agent specifying the Core version used by the site, as well as any additional agents specified in the request object. If an additional agent has a falsy value for the version, it will not be included in the header. Add unit tests to confirm this behavior.

Because we are importing package.json to get the version, in order to get the `commonjs` and `esm` imports to work correctly, the `outDir` was changed from `lib` to `dist`.

J=SLAP-1947
TEST=auto, manual

Check that Jest tests pass and see that the correct header is passed when using the test-site.


src/infra/AutocompleteServiceImpl.ts
```diff
@@ -70,8 +71,17 @@ export class AutocompleteServiceImpl implements AutocompleteService {
     };
 
     const response = 'token' in this.config
-      ? await this.httpService.get<ApiResponse>(this.universalEndpoint, queryParams, this.config.token)
-      : await this.httpService.get<ApiResponse>(this.universalEndpoint, queryParams);
+      ? await this.httpService.get<ApiResponse>(
```
oshi97, line 74: Is it possible to remove this ternary op? I know I originally removed the ternary op but not sure if I missed something wrong with doing so?

src/models/core/ClientSdk.ts
```diff
@@ -0,0 +1,10 @@
+
+/**
+ * Additional agents to add to the Client-SDK HTTP header.
+ *
+ * @public
+ **/
+export type CustomClientSdk = Record<string, string> & { ANSWERS_CORE?: never };
+
+/** The Client-SDK including the ANSWERS_CORE agent. */
+export type ClientSdk = Record<string, string> & { ANSWERS_CORE: string };
```
cea2aj, line 10: Can we write these as interfaces instead? In Typescript, interfaces are generally preferred because they are easier to extend.

src/infra/AutocompleteServiceImpl.ts
nmanu1: I added the ternary operator back in because otherwise we have to check for `undefined` (or `expect.anything()`) as the `authToken` param for each unit test that doesn't specify a token

oshi97: ah I see, either is fine then! I was mostly wondering if the behavior didn't work as expected

src/models/core/ClientSdk.ts
nmanu1: Sure, I can write them as interfaces instead. But, I think `CustomClientSdk` would have to be something like this, where the additional agents could be specified as `undefined` to allow for an optional `ANSWERS_CORE` property:
``` 
interface CustomClientSdk {
  [agent: string]: string | undefined,
  ANSWERS_CORE?: never
}
```

etc/answers-core.api.md
```diff
@@ -96,6 +96,11 @@ export interface CombinedFilter {
 // @public
 export type Context = any;
 
+// @public
+export type CustomClientSdk = Record<string, string> & {
```
cea2aj, line 100: Did the spec say to call it CustomClientSdk? I tried looking for the spec but I couldn't find it

src/models/core/ClientSdk.ts
cea2aj: In some cases defining a type as an interface adds no additional complexity. In those cases, I think we should always pick an interface. Sometimes, using interfaces introduces some complexity that wouldn't be present in types. In those cases, it can go either way depending on if the documentation or simplicity in the code is more important. I think I'm leaning towards using an interface, but up to you

tsconfig.base.json
```diff
@@ -7,7 +7,8 @@
     "forceConsistentCasingInFileNames": true,
     "declaration": true,
     "declarationMap": false,
-    "sourceMap": true
+    "sourceMap": true,
+    "resolveJsonModule": true
```
cea2aj, line 11: Our public interface allows people to import `@yext/answers-core/lib/commonjs` or `@yext/answers-core/lib/esm`. When we import the package.json file from the src files, the output files in the lib directories are restructured since the src files need to be able to import the package.json file. This means that the imports listed above will break since there is no longer an index.js file in those folders. The package.json files will be read instead, and the `main` paths in the package.json files of those folders will be invalid.

One solution is to update the outDir of the tsconfig.esm.json and the tsconfig.commonjs.json files to be dist/esm and dist/commonjs. Then we can check-in the lib folder into the repo and add small package.json files which point to the corresponding index and bundle.d.ts files in dist. You can see an example of this with legacy/package.json, however we should use the `main` and `module` keys instead of `browser` to point to the code.

src/models/core/ClientSdk.ts
nmanu1: It didn't add too much complexity, so I changed them to interfaces

src/infra/AutocompleteServiceImpl.ts
```diff
@@ -70,8 +71,17 @@ export class AutocompleteServiceImpl implements AutocompleteService {
...
+        this.universalEndpoint,
+        queryParams,
+        getClientSdk(request.customClientSdk),
+        this.config.token
```
tmeyer2115, line 78: Here and below the indentation needs to be corrected. The function args should be further in by one tab, they should not be flush with `await`.

src/infra/HttpServiceImpl.ts
```diff
@@ -22,13 +22,17 @@ export class HttpServiceImpl implements HttpService {
...
     authToken?: string,
   ): Promise<T> {
-    return this.fetch(url, queryParams, {
+    return fetch(url, queryParams, {
       method: RequestMethods.GET,
       mode: 'cors',
       credentials: 'include',
-      ...(authToken && { headers: { Authorization: `Bearer ${authToken}` }}),
+      headers: {
+        'Client-SDK': formatAsHttpHeader(clientSdk),
```
tmeyer2115, line 33: In the spec, the header is all caps (`CLIENT-SDK`) does this header get capitalized automatically somewhere?

src/infra/HttpServiceImpl.ts
```diff
@@ -39,34 +43,46 @@ export class HttpServiceImpl implements HttpService {
...
+  reqInit: RequestInit
+): Promise<Response> {
+  const urlWithParams = addParamsToURL(url, queryParams);
+  if (typeof(window) !== 'undefined' && window.fetch) {
+    return window.fetch(urlWithParams, reqInit);
   }
+  return crossFetch(urlWithParams, reqInit);
 }
+
+function formatAsHttpHeader(jsonHeader: Record<string, string>) {
```
tmeyer2115, line 80: Can we add some docs here? Also, small nit: it feels like what this method is doing is serializing a json object into an HTTP header. I think the name `jsonHeader` is a bit misleading. We're not converting one header into another. We're formatting JSON data into a header.

src/models/core/ClientSdk.ts
```diff
@@ -0,0 +1,21 @@
+
+/**
+ * Additional agents to add to the Client-SDK HTTP header.
+ *
+ * @public
+ **/
+export interface CustomClientSdk {
```
tmeyer2115, line 7: I think we should consider renaming this `interface` and the one below. The name is closely tethered to an internal implementation detail (the name of the header we use). But, that's not the level of abstraction a user operates on. They're operate at a higher level of abstraction. They need to know that this is a way for them to add other Answers Agents and have them be passed on LiveAPI calls. I don't think the name `ClientSdk` or `CustomClientSdk` will be very intuitive for them.

src/infra/HttpServiceImpl.ts
oshi97: http headers are case insensitive! if you view them in chrome dev tools they all get turned to lowercase. That being said maybe matching the spec is still nicer here?

src/models/core/ClientSdk.ts
nmanu1: What do you think of `AnswersAgents` and `CustomAnswersAgents` instead?

src/infra/HttpServiceImpl.ts
oshi97: we could still update though

src/models/core/ClientSdk.ts
oshi97: I actually feel like we should try to be relatively consistent with the API here, since I think these headers end up getting surfaced in the data hub as CLIENT_SDK (or something similar?)

Plus the only people that are going to even use this interface are library maintainers who want to use core in their own libraries, it's not your average joe who just wants to build an answers experience.

src/models/autocompleteservice/AutocompleteRequest.ts
```diff
@@ -8,6 +10,8 @@ export interface UniversalAutocompleteRequest {
   input: string,
   /** Enables session tracking. */
   sessionTrackingEnabled?: boolean,
+  /** {@inheritDoc CustomClientSdk} */
+  customClientSdk?: CustomClientSdk
```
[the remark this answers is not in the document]
nmanu1: Typescript already has a `Request` interface, so I added a base `AnswersRequest` interface

src/infra/HttpServiceImpl.ts
oshi97: `Client-SDK` is what is used in the slack discussions so I'm not sure which is the correct version

src/models/core/ClientSdk.ts
oshi97: I agree the name is unintuitive... but I think it's better to just have it be consistent all the way through here (maybe Rose can weigh in more on this though). That being said I don't think our interfaces have to be named exactly `ClientSdk`  or `CustomClientSdk` but they should somewhat resemble the HTTP header?

tsconfig.cjs.json
```diff
@@ -3,6 +3,6 @@
   "compilerOptions": {
     "target": "es2015",
     "module": "commonjs",
-    "outDir": "lib/commonjs"
+    "outDir": "dist/commonjs"
```
oshi97, line 6: is the point of this change so that we can get rid of the setup-lib-types script?

nmanu1: this is so we can just have package.json files in `lib/commonjs` and `lib/esm` that point to the correct index.js file for each. otherwise, the package.json was copied in from the top-level one and the `main` and `module` paths would be incorrect, causing `@yext/answers-core/lib/commonjs` and `@yext/answers-core/lib/esm` imports to fail. removing the script is just a side effect as it's no longer necessary

src/infra/HttpServiceImpl.ts
tmeyer2115: Gotcha! Sounds like it doesn't matter either way.

src/models/core/ClientSdk.ts
tmeyer2115: If `CLIENT_SDK` is the key in the data hub, then I would agree with you Oshi. I do think we could improve the docs a bit though.

src/infra/HttpServiceImpl.ts
nmanu1: Sure, I renamed `jsonHeader` to `clientSdk` and added docs with an example that should make it more clear.

src/models/core/ClientSdk.ts
nmanu1: I updated the docs for `CustomClientSdk`. Let me know if more clarification is needed

src/infra/HttpServiceImpl.ts
```diff
@@ -39,34 +43,61 @@ export class HttpServiceImpl implements HttpService {
...
+): Promise<Response> {
+  const urlWithParams = addParamsToURL(url, queryParams);
+  if (typeof(window) !== 'undefined' && window.fetch) {
+    return window.fetch(urlWithParams, reqInit);
   }
+  return crossFetch(urlWithParams, reqInit);
 }
+
+/**
+ * Converts the JSON representing the Client-SDK agents into the format of an HTTP header.
```
oshi97, line 81: instead of `the format of an HTTP header` I would maybe say `the expected HTTP header format`?
I don't think there's anything "wrong" with putting json in an HTTP header it's just not conventional (afaik, I'm no http expert)

src/infra/HttpServiceImpl.ts
```diff
@@ -39,34 +43,61 @@ export class HttpServiceImpl implements HttpService {
...
+ * Agents that have an undefined version are excluded from the returned header.
```
oshi97, line 82: Here we don't just exclude `undefined` values but *any* falsy value, i.e. null and blank string as well
Do we even let people have undefined values though? The type for `clientSdk` is `Record<string, string>` not `Record<string, string | undefined>`

src/models/core/ClientSdk.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Additional agents and their versions used to create the Answers experience. The information for these
+ * agents is added to the Client-SDK HTTP header along with that of the ANSWERS_CORE agent.
+ *
+ * @public
+ **/
+export interface CustomClientSdk {
+  /** A mapping of the additional agents that are part of the Client-SDK to their versions. */
+  [agent: string]: string | undefined,
+  /**
+   * The ANSWERS_CORE agent should not be supplied. Instead, it will be automatically added to the
```
oshi97, line 12: The docs read really nicely!

src/infra/HttpServiceImpl.ts
nmanu1: We let them specify undefined values in CustomClientSdk and it seems like they are able to persist without any type check problems in objects of type `ClientSDK` or `Record<string, string>`

oshi97: after updating the docs lgtm!

oshi97 (approved):

src/infra/HttpServiceImpl.ts
nmanu1: I removed this part of the docs because I moved the check for a falsy value to `getClientSdk`. This way, when we get an object of type ClientSdk, it won't have unexpected `undefined` values either

PR #144, opened 2022-03-16
update CustomSdkClients to AdditionalHttpHeaders

oshi97:
This changes both the name and shape of CustomSdkClients.

J=SLAP-1971
TEST=manual,auto

see that the test site passes custom and built in Client-SDK header
values


etc/answers-core.api.md
```diff
@@ -4,6 +4,12 @@
 
 ```ts
 
+// @public
+export interface AdditionalHttpHeaderValues {
+    // (undocumented)
```
nmanu1, line 9: nit: missing a doc

src/infra/QuestionSubmissionServiceImpl.ts
```diff
@@ -54,14 +54,14 @@ export class QuestionSubmissionServiceImpl implements QuestionSubmissionService
         this.endpoint,
         queryParams,
         body,
-        getClientSdk(request.customClientSdk),
+        getClientSdk(request.additionalHttpHeaderValues),
         this.config.token
       )
       : await this.httpService.post<ApiResponse>(
```
yen-tt, line 60: since we mostly keep get/post methods to be pretty generic, would it make sense to pass something like a headerParams or additionalHeaderParams which includes `Client-SDK` instead of passing it directly in as an argument?

src/services/HttpService.ts
```diff
@@ -8,15 +8,15 @@ export interface HttpService {
...
+    clientSdk: Record<string, string>,
     authToken?: string
   ): Promise<T>;
 
   post<T>(
     url: string,
     queryParams: QueryParams,
     body: QueryParams,
-    clientSdk: ClientSdk,
+    clientSdk: Record<string, string>,
```
yen-tt, line 19: since we mostly keep get/post methods to be pretty generic, would it make sense to pass something like a headerParams or additionalHeaderParams which includes `Client-SDK` instead of passing it directly in as an argument?

src/utils/getClientSdk.ts
```diff
@@ -1,12 +1,18 @@
-import { ClientSdk, CustomClientSdk } from '../models/core/ClientSdk';
+import { AdditionalHttpHeaderValues } from '../models/core/AdditionalHttpHeaderValues';
 import packageJson from '../../package.json';
 
 const { version } = packageJson;
 
-export function getClientSdk(customClientSdk?: CustomClientSdk): ClientSdk {
+export function getClientSdk(
+  additionalHttpHeaderValues?: AdditionalHttpHeaderValues
+): Record<string, string> {
```
yen-tt, line 8: can the return value type be `ClientSDKHeaderValues`?

src/models/core/AdditionalHttpHeaderValues.ts
```diff
@@ -1,11 +1,20 @@
 
+/**
+ * AdditionalHttpHeaderValues allows users to specify additional values for specific HTTP headers.
+ *
+ * @public
+ **/
+export interface AdditionalHttpHeaderValues {
```
nmanu1, line 7: do we anticipate there being more headers added in the future? otherwise I'm not sure how much value this interface adds

src/utils/getClientSdk.ts
nmanu1: I don't think that would work because `ClientSDKHeaderValues` specifically prevents the Core header from being specified, which is being added in this function

oshi97: no because we specify ANSWERS_CORE here. I thought about it for a sec and I don't think we need an internal type here since all HttpServiceImpl cares about is that it's a Record<string, string>. I think having an extra type complicates things slightly

src/models/core/AdditionalHttpHeaderValues.ts
oshi97: Yes this is what Rose decided on, I raised the issue when she first had this idea, and then Tom also brought it up. I probably should have just written a comment saying we talked about this

src/services/HttpService.ts
oshi97: If you remember we ran into an issue where we weren't passing cookies in requests. After that, I realized that we were trying to make HttpService too generic, more generic than it needed to be. I think it will be very easy to refactor this to handle multiple headers if we need to, but that's a long way off. This also prevents people from cheating additional headers into the request by using the library through javascript, since we won't be doing a loop over the additional header values but just picking exactly the one we want, Client-SDK.

src/models/core/AdditionalHttpHeaderValues.ts
nmanu1: I see. So would this go all the way up to the Theme and Headless-React? Like we would want those libraries (and therefore the SDK and Headless) to also be able to specify other headers in the future instead of just passing in the Client-SDK header for now?

etc/answers-core.api.md
oshi97: updated!

src/models/core/AdditionalHttpHeaderValues.ts
oshi97: Yes all the other libraries will have this updated config shape
The theme and headless-react will not expose this to the user though, at least not yet

src/services/HttpService.ts
oshi97: Since you might be curious, here's the PR in question https://github.com/yext/answers-core/pull/97

yen-tt (approved):

src/models/core/AdditionalHttpHeaders.ts
```diff
@@ -1,11 +1,21 @@
...
+
 /**
  * Additional agents and their versions used to create the Answers experience. The information for these
  * agents is added to the Client-SDK HTTP header along with that of the ANSWERS_CORE agent.
  *
  * @public
- **/
-export interface CustomClientSdk {
+ */
+export interface ClientSDKHeaderValues {
```
yen-tt, line 18: if we are dropping `Values` from AdditionalHttpHeaders, do we want to drop it here as well, and just have `ClientSDKHeader`?

oshi97: Here it actually is values though

PR #145, opened 2022-03-22
support number range facet/filter

yen-tt:
This pr contains changes to support number range facets/filters. From discussions with backend, the number range filter would return in a format like this:
```
"c_price": {
          "$between": {
              "$ge": 4.0,
              "$lt": 800.0
       }
 }
```

updated enum Matcher to include a new matcher `$between` and added a new interface `NumberRangeValue` to represent the value of "$between" matcher. The alternative approach for `NumberRangeValue` is to make a type like this:
`type NumberRangeValue = { [matcher in NumberFacetMatcher]: string; };`
But I think the current approach of `NumberRangeValue` interface with start and end field provides a clearer restriction to what the value can look like and it's also more align to the configuration interface that product want in the spec. But this does mean some transformation is needed in between requests and response. 

Updated update shapeFilterForApi to transform the number range facets/filters properly for api requests.
Updated createFilter to transform facets/filters properly when constructing core response.

J=SLAP-1977
TEST=auto

see that jest tests passed


etc/answers-core.api.md
```diff
@@ -136,17 +136,14 @@ export enum Direction {
 // @public
 export interface DisplayableFacet extends Facet {
     displayName: string;
-    fieldId: string;
```
cea2aj, line 139: We decided to include properties from interfaces which are being extended in order to provide slightly more readable documentation.  Is there a need to remove the duplicated props here and in DisplayableFacetOption?

etc/answers-core.api.md
```diff
@@ -287,6 +285,21 @@ export interface NearFilterValue {
...
+    value: number;
+}
+
+// @public
+export type NumberRangeMatcher = Matcher.GreaterThan | Matcher.GreaterThanOrEqualTo | Matcher.LessThan | Matcher.LessThanOrEqualTo;
+
+// @public
+export interface NumberRangeValue {
+    end: NumberRangeLimit;
+    start: NumberRangeLimit;
```
cea2aj, line 300: What do you think of making these optional? The spec says that if an 'end' or a 'start' isn't specified, the range is unbounded

etc/answers-core.api.md
```diff
@@ -287,6 +285,21 @@ export interface NearFilterValue {
...
+    matcher: NumberRangeMatcher;
+    value: number;
+}
+
+// @public
+export type NumberRangeMatcher = Matcher.GreaterThan | Matcher.GreaterThanOrEqualTo | Matcher.LessThan | Matcher.LessThanOrEqualTo;
+
+// @public
+export interface NumberRangeValue {
+    end: NumberRangeLimit;
```
cea2aj, line 299: What are your thoughts on creating a `LowerNumberRange` and a `UpperNumberRange` model? That way we can statically enforce that nobody creates a `NumberRangeValue` with two `GreaterThan` or two `LessThan` matchers

src/serializers/serializeStaticFilters.ts
```diff
@@ -25,11 +26,22 @@ function shapeCombinedFilterForApi(combinedFilter: CombinedFilter): StaticFilter
     : { [combinedFilter.combinator]: shapedFilters };
 }
 
+export function isNumberRangeValue(data: unknown): data is NumberRangeValue {
```
cea2aj, line 29: Can we move `isNumberRangeValue` next to the `NumberRangeValue` model? I don't expect it to change much, but it's really import that type guards accurately reflect the model, because if they differ, runtime errors can creep in.

src/serializers/serializeStaticFilters.ts
```diff
@@ -1,6 +1,7 @@
+import { NumberRangeValue } from '../models/searchservice/common/NumberRangeValue';
 import { CombinedFilter } from '../models/searchservice/request/CombinedFilter';
 import { Filter } from '../models/searchservice/request/Filter';
-import { StaticFilters } from '../models/searchservice/request/StaticFilters';
+import { StaticFilters, Filter as InternalFilter } from '../models/searchservice/request/StaticFilters';
```
cea2aj, line 4: Thoughts on `ApiFilter`? We could even rename it in the StaticFilters file since it's not part of our public API. Having `ApiStaticFilters` could be helpful too.

src/transformers/core/createFilter.ts
```diff
@@ -1,13 +1,23 @@
...
 
 export function createFilter(filter: any): Filter {
   const fieldId = Object.keys(filter)[0];
   const matcher = Object.keys(filter[fieldId])[0];
-
+  const rawValue = filter[fieldId][matcher];
+  let value: string | number | boolean | NearFilterValue | NumberRangeValue;
+  if (matcher === Matcher.Between) {
+    value = {
+      start: { matcher: Object.keys(rawValue)[0], value: Object.values(rawValue)[0] },
```
cea2aj, line 12: Is the first key guaranteed to be the start matcher? Even so, it's probably a good idea to some checking so that we aren't relying on key ordering

tests/infra/SearchServiceImpl.ts
```diff
@@ -341,6 +341,44 @@ describe('SearchService', () => {
...
+                matcher: Matcher.Between,
+                value: {
+                  start: { matcher: Matcher.GreaterThan, value: 0 },
+                  end: { matcher: Matcher.LessThan, value: 10 }
+                }
+              },
+              {
+                matcher: Matcher.Between,
+                value: {
+                  start: {matcher: Matcher.GreaterThanOrEqualTo, value: 30 },
```
cea2aj, line 363: nit: spacing here and in serializeFacets

etc/answers-core.api.md
yen-tt: I see. I noticed the duplications and decided to remove it, didn't realize it was intentional for documentation purpose. I will add it back in!

src/serializers/serializeStaticFilters.ts
yen-tt: I originally thought we may not want functions in models folder, and keep it for interfaces/types stuff only. But I remember we did have some type guard functions in models folder for answers-react-components as well. I will move it

etc/answers-core.api.md
yen-tt: From the first example facet in spec, it looks like it's always bounded in the backend. However, after discussing with product, the backend will handle the case when one of the limit is not specified. I will update this

src/transformers/core/createFilter.ts
```diff
@@ -20,4 +16,16 @@ export function createFilter(filter: any): Filter {
     matcher: matcher as Matcher,
     value
   };
+}
+
+function createNumberRangeValue(value: any): NumberRangeValue {
+  const numberRangeValue: NumberRangeValue = {};
+  Object.keys(value).forEach(limitMatcher => {
+    if (limitMatcher === Matcher.GreaterThan || limitMatcher === Matcher.GreaterThanOrEqualTo) {
```
nmanu1, line 24: can we use the Upper/Lower NumberRangeMatchers here?

src/models/searchservice/request/StaticFilters.ts
```diff
@@ -5,15 +5,15 @@ import { NearFilterValue } from './Filter';
...
+  [combinator: string]: (ApiStaticFilters | ApiFilter)[] | ApiFilter | ApiStaticFilters;
 }
 
 /**
  * Represents a single filter.
  *
  * @internal
  */
-export interface Filter {
+export interface ApiFilter {
```
oshi97, line 17: Can we update the comment to say how this is different than a regular Filter?

src/transformers/core/createFilter.ts
yen-tt: the NumberRangeMatchers are types so they can't be use here in the conditional logics as values

src/models/searchservice/common/NumberRangeValue.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @public
+ */
+export interface NumberRangeValue {
+  /** Start limit of the number range value. */
+  start?: LowerNumberRangeLimit
+  /** End limit of the number range value. */
+  end?: UpperNumberRangeLimit
+}
+
+export function isNumberRangeValue(data: unknown): data is NumberRangeValue {
```
oshi97, line 54: I think we can put this type guard inside serializeStaticFilters? I kind of like putting it here centralizes it, but we also normally don't put non type/interface code inside the models folder. We're only using it in a single place right now, and maybe we can centralize if we need it later?

yen-tt: It was originally in `serializeStaticFilters` but connor suggested it to move to here so it's easier to be consistent with NumberRangeValue if there were any changes to it (refer to the comment made above). I don't have a strong preference here

src/models/searchservice/common/NumberRangeValue.ts
```diff
@@ -0,0 +1,56 @@
+import { Matcher } from './Matcher';
+
+/**
+ * Matcher values of a lower limit for a filter with a $between {@link Matcher}.
+ *
+ * @public
+ */
+export type LowerNumberRangeMatcher = Matcher.GreaterThan | Matcher.GreaterThanOrEqualTo;
```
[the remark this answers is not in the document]
yen-tt: Fair, I will put them directly in `NumberRangeValue `

src/serializers/serializeFacets.ts
```diff
@@ -8,10 +8,10 @@ export function serializeFacets(filters: Facet[]): string {
     const shapedFacets = shapeFacetOptionArrayForApi(facet.options, fieldId);
     obj[fieldId] = shapedFacets;
     return obj;
-  }, {} as { [key: string]: StaticFilters[] }));
+  }, {} as { [key: string]: ApiStaticFilters[] }));
```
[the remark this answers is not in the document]
yen-tt: yep!

src/models/searchservice/common/NumberRangeValue.ts
oshi97: Oh omg I missed the comment. We can keep it! The placement isn't crucial

src/serializers/serializeFacets.ts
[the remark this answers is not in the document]
oshi97: I think this should work? though the type assert is probably fine if you don't want to change
```ts
  return JSON.stringify(filters.reduce<Record<string, ApiStaticFilters[]>>((obj, facet) => {
    const fieldId = facet.fieldId;
    const shapedFacets = shapeFacetOptionArrayForApi(facet.options, fieldId);
    obj[fieldId] = shapedFacets;
    return obj;
  }, {}));`
```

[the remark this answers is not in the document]
oshi97: I would probably use a `Record` regardless just for simplicity

oshi97 (approved):
lgtm!

PR #148, opened 2022-03-24
use shared eslint

oshi97:
manually fixed 2 non auto-fixable issues in HttpServiceImpl with the tsdoc
all other issues were auto-fixed

3 license issues are introduced (through `argparse`, `axe-core`, and `language-subtag-registry`), but they are all on the approved licenses spreadsheet and are fine with legal since we have added the license preamble for the `language-subtag-registry` in this PR.

J=SLAP-2003, SLAP-2064
TEST=build


nmanu1 (commented):
it looks like a Snyk check is failing?

package.json
```diff
@@ -47,11 +47,15 @@
     "@microsoft/api-documenter": "^7.13.76",
     "@microsoft/api-extractor": "^7.19.2",
     "@types/jest": "^26.0.15",
-    "@typescript-eslint/eslint-plugin": "^4.6.0",
-    "@typescript-eslint/parser": "^4.6.0",
+    "@typescript-eslint/eslint-plugin": "^5.16.0",
+    "@typescript-eslint/parser": "^5.16.0",
```
tmeyer2115, line 51: Do we still need these as top-level dependencies? I'm wondering if the shared ESLint config dep takes care of them.

nmanu1: I think Oliver tried removing them and it caused tests to fail, but I'll look into it more to make sure

nmanu1: We should be able to remove all of these dependencies because npm 7 and later automatically installs peer dependencies. But, we run tests using node 12 and 14, which use npm 6, so the peer dependencies aren't installed and those tests fail. So, if we want to continue testing on the older node versions, I think these need to be specified as top-level dependencies

tmeyer2115 (approved):

PR #149, opened 2022-04-05
Update defaultApiVersion

nmanu1:
Update the `defaultApiVersion` to match the v param for the new LiveApi version that will support number range facets.

J=SLAP-2024
TEST=auto, manual

Check that updated Jest tests pass. Use the Core test-site and see that the updated v param is passed with each request and all endpoints (except for question submission) return a 200 status.

Once the backend work was done, performed a full end-to-end test and saw that the component lib test-site using Core locally passed the new v param and had number range facets returned in the response. The same site using Core from NPM passed the old v param and did not have number range facets returned.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@yext/answers-core",
-      "version": "1.6.0-beta.5",
+      "version": "1.6.0-beta.6",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@babel/runtime-corejs3": "^7.12.5",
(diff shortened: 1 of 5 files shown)


oshi97 (approved):

cea2aj (approved):

nmanu1:
I will hold off on merging this in until the backend work is done and we can do a full end-to-end test of the functionality.

nmanu1:
> I will hold off on merging this in until the backend work is done and we can do a full end-to-end test of the functionality.

Update: the end-to-end test was successful

PR #152, opened 2022-04-12
Bump version to v1.6.0-beta.8

nmanu1:
Bumps the beta version to incorporate the changes from reverting the `FilterSearchResponse` updates (#112). These updates will not be part of v1.6.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yext/answers-core",
-  "version": "1.6.0-beta.7",
+  "version": "1.6.0-beta.8",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@yext/answers-core",
-      "version": "1.6.0-beta.7",
+      "version": "1.6.0-beta.8",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@babel/runtime-corejs3": "^7.12.5",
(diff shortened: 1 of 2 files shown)


oshi97:
I think we updated the SDK to make use of these changes, do those also need to be reverted?

nmanu1:
> I think we updated the SDK to make use of these changes, do those also need to be reverted?

Yes, the SDK [changes](https://github.com/yext/answers-search-ui/pull/1599) are in develop, so when I make the release branch, I'll revert that commit as well

yen-tt:
> > I think we updated the SDK to make use of these changes, do those also need to be reverted?
> 
> Yes, the SDK [changes](https://github.com/yext/answers-search-ui/pull/1599) are in develop, so when I make the release branch in the SDK, I'll revert that commit as well

I believe previously we wanted to wait for a new SDK minor for this change. Since we are releasing a v1.14, I think this can be included? Should probably ask tom/product on this

nmanu1:
> > > I think we updated the SDK to make use of these changes, do those also need to be reverted?
> > 
> > 
> > Yes, the SDK [changes](https://github.com/yext/answers-search-ui/pull/1599) are in develop, so when I make the release branch in the SDK, I'll revert that commit as well
> 
> I believe previously we wanted to wait for a new SDK minor for this change. Since we are releasing a v1.14, I think this can be included? Should probably ask tom/product on this

We hadn't been asked to merge this in originally (which is why I had reverted it in the release branch), but after talking to Rose about it more, we will be including it, so this version bump and the revert is no longer needed

PR #154, opened 2022-04-21
Use reusable workflows

nmanu1:
Update the github workflows to use the callable workflows in the `slapshot-reusable-workflows` repo.

J=SLAP-2005
TEST=auto

- See that the `coverage`, `update_docs`, and `run_tests` workflows run on this PR as expected.
- Test on another branch (`dev/test-reusable-workflows`) and see that docs are updated when needed.
- Fork the repo and commit to master. See that the `sync_develop_and_main` workflow creates a PR as expected.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -10,6 +10,6 @@ on:
 
 jobs:
   call_code_coverage:
-    uses: yext/slapshot-reusable-workflows/.github/workflows/coverage.yml@main
+    uses: yext/slapshot-reusable-workflows/.github/workflows/coverage.yml@v1
     secrets:
       caller_github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/coverage.yml
```diff
@@ -9,22 +9,7 @@ on:
...
-        node-version: ${{ matrix.node-version }}
-    - run: npm ci
-    - run: npx jest --coverage
-    - name: Coveralls
-      uses: coverallsapp/github-action@master
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+  call_code_coverage:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/coverage.yml@main
```
cea2aj, line 13: Do you think it would be a good idea to point at a versioned branch rather than main? For example, we could make a branch called 'v1' and push and minor or patch versions to it as needed. That way if we ever want to make breaking changes, we can make a 'v2' branch off of master without breaking the repos which depend on the 'v1' branch

nmanu1: Yeah, that makes sense. I'll update

cea2aj (approved):

PR #158, opened 2022-05-09
Update GENERIC to CUSTOM_SEARCHER for result's Source field

yen-tt:
Currently, the core library describes the source of results as 'Generic' if it does not match one of the defined search backends (e.g. Google, Bing, Zendesk, etc). Confirmed with Watson that the result will supplied one of the 6 enum values: KNOWLEDGE_MANAGER, CUSTOM_SEARCHER, GOOGLE_CSE, BING_CSE, ZENDESK, ALGOLIA.

This pr updates GENERIC to CUSTOM_SEARCHER source type instead. 

J=SLAP-1213
TEST=manual

jest tests passed

### DIFF
--- a/src/models/searchservice/response/Result.ts
+++ b/src/models/searchservice/response/Result.ts
@@ -10,7 +10,7 @@ export interface Result {
   /** Raw entity profile data in the shape of key-value pairs. */
   rawData: Record<string, unknown>,
   /** {@inheritDoc Source}*/
-  source: Source,
+  source: Source | string,
   /** The index of the result among the other results in the search. */
   index?: number,
   /** The name of the result. */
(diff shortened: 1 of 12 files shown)


src/models/searchservice/response/Result.ts
```diff
@@ -10,7 +10,7 @@ export interface Result {
   /** Raw entity profile data in the shape of key-value pairs. */
   rawData: Record<string, unknown>,
   /** {@inheritDoc Source}*/
-  source: Source,
+  source: Source | string,
```
oshi97, line 13: Could we restrict `source` to the specific strings in the enum instead of any string?

yen-tt: Looks like backend may return custom strings that's outside of the defined set of enums we currently have. I will confirm with watson if it can be any string or if there's other defined strings we didn't account for

yen-tt: Good call! Confirmed with Watson that it will always be one of the 6 enum values specified in this pr's description. So we only need to update GENERIC to CUSTOM_SEARCHER enum.

oshi97 (approved):

PR #159, opened 2022-05-09
Support AppliedQueryFilter's type and details field

yen-tt:
update AppliedQueryFilter interface and added new interfaces to follow the vertical query response model from [liveAPI](https://hitchhikers.yext.com/docs/liveapis/answersapi/verticalsearch/). Confirmed with Watson that:
1) `details` field only exist when the filter type is `PLACE`
2) `details` field always contains `featureTypes`, `latitude`, `longitude`, and `placeName`. `boundingBox` field is only sometimes (such as when the the location is of `locality`/`region` feature specific, instead of `address` specific)

J=SLAP-1259
TEST=manual

jest tests passed


src/transformers/searchservice/createAppliedQueryFilter.ts
```diff
@@ -5,6 +5,8 @@ export function createAppliedQueryFilter(data: any): AppliedQueryFilter {
   return {
     displayKey: data.displayKey,
     displayValue: data.displayValue,
-    filter: createFilter(data.filter)
+    filter: createFilter(data.filter),
+    type: data.type,
+    ...(data.details && { details: data.details })
```
oshi97, line 10: I guess the plain `details: data.details` doesn't work here?

oshi97: is it because data.details could be a null or falsy value that's not undefined?

oshi97 (approved):

src/transformers/searchservice/createAppliedQueryFilter.ts
yen-tt: if data.details is not provided by liveAPI, I thought it would be best to leave `details` field out of the returned object instead of having `details: undefined` in the object. But if that doesn't matter, I could do the plain `details: data.details`

PR #160, opened 2022-05-10
Merge develop into release/v1.6

nmanu1:


tmeyer2115:
I believe the security issues pointed out by Snyk can all be ignored. I think we've seen them before on other PRs and they're a non-issue. Can you quickly confirm though?

yen-tt (approved):
I see that the snyk errors is from the 3 license issues are introduced (through argparse, axe-core, and language-subtag-registry) in #148 . If they are resolved, we should temporarily ignore it (for 90 days)

PR #164, opened 2022-05-11
Support failedVerticals

nmanu1:
Add support for `failedVerticals` returned in the response for universal queries.

J=SLAP-1214
TEST=auto, manual

Check that the new and existing Jest tests pass. Use the Core test-site and see that responses with and without failed verticals are handled correctly.


src/transformers/searchservice/createUniversalSearchResponse.ts
```diff
@@ -17,6 +18,7 @@ export function createUniversalSearchResponse(data: any): UniversalSearchRespons
     spellCheck: data.response.spellCheck && createSpellCheck(data.response.spellCheck),
     locationBias: data.response.locationBias && createLocationBias(data.response.locationBias),
     uuid: data.meta.uuid,
-    queryRulesActionsData: data.response.queryRulesActionsData
+    queryRulesActionsData: data.response.queryRulesActionsData,
+    failedVerticals: data.response.failedVerticals && data.response.failedVerticals.map(createFailedVertical)
```
yen-tt, line 22: I noticed we have `failedVerticals` field in our autocomplete response fixtures, can you verify if changes are needed for creating universal autocomplete response as well or update our fixtures if it's not suppose to be there?

nmanu1: Yes, `failedVerticals` is also returned by the filter search response. But, we handle that in `createFilterSearchResponse` and throw an error if it's populated. So, the fixture is for testing that error behavior, and I don't think changes to the response are needed

yen-tt: ah I see, is there a reason why we are returning failedVerticals in one response but throwing an error in another?

nmanu1: I think it definitely makes sense to include it in the response for universal search because other verticals could still return results. for filter search, it was initially added in the response too, but based on [this comment](https://github.com/yext/answers-core/pull/112#discussion_r746928819), it was changed to throw an error instead

src/models/searchservice/response/FailedVertical.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @public
+ */
+export interface FailedVertical {
+  /** The vertical key associated with the failed vertical. */
+  verticalKey: string,
+  /**
+   * The type of error causing the failure. Possible options are
+   * "TIMEOUT", "BACKEND_ERROR", "INVALID_CONFIG", and "INVALID_QUERY".
+   */
+  errorType: string,
```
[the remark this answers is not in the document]
nmanu1: updated!

PR #165, opened 2022-05-25
add SandboxEndpoints

oshi97:
This PR exports a new SandboxEndpoints object which contains our
liveapi sandbox endpoint urls.
As part of this, our various api services were updated to no longer default
endpoints within the service implementation, but instead do the defaulting
within provideCore.

Moved provideCore to its own file to separate concerns with index.ts

The documentation doesn't say exactly what the endpoints are, if we wanted we
removing the Required<Endpoints> type and keeping the as const would let the URLs
show up in documentation, but I would rather keep the Required type.

J=SLAP-2106
TEST=manual,auto

update test-site to use SandboxEndpoints, see it work as expected


src/provideCore.ts
```diff
@@ -0,0 +1,32 @@
+import { SearchServiceImpl } from './infra/SearchServiceImpl';
```
oshi97, line 1: this code was not changed at all just moved to a different file

src/constants.ts
```diff
@@ -10,4 +10,19 @@ export const defaultEndpoints: Required<Endpoints> = {
   universalAutocomplete: 'https://liveapi-cached.yext.com/v2/accounts/me/answers/autocomplete',
   verticalAutocomplete: 'https://liveapi-cached.yext.com/v2/accounts/me/answers/vertical/autocomplete',
   filterSearch: 'https://liveapi-cached.yext.com/v2/accounts/me/answers/filtersearch',
-};
\ No newline at end of file
+} as const;
```
nmanu1, line 13: for my own understanding, why is the const assertion needed?

src/constants.ts
```diff
@@ -10,4 +10,19 @@ export const defaultEndpoints: Required<Endpoints> = {
...
+
+/**
+ * The endpoints to use for sandbox experiences
```
nmanu1, line 16: nit: missing period

src/constants.ts
oshi97: this marks all the properties as read only

oshi97: it also keeps the type as narrow as possible, though that doesn't matter here
here's a good post on it https://blog.logrocket.com/const-assertions-are-the-killer-new-typescript-feature-b73451f35802/

package.json
```diff
@@ -15,7 +15,7 @@
     "node": ">=12"
   },
   "scripts": {
-    "test": "npm run lint && jest",
+    "test": "npm run lint && jest --collectCoverage",
```
tmeyer2115, line 18: Was `collectCoverage: true` not working properly?

tests/infra/AutocompleteServiceImpl.ts
```diff
@@ -306,3 +307,64 @@ describe('additionalQueryParams are passed through', () => {
     }));
   });
 });
+
+describe('sandbox endpoints work as expected', () => {
```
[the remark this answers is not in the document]
tmeyer2115: Personally, I lean to 1.

package.json
oshi97: It's a little annoying to have to specify not to collect coverage during the development loop for a test, though this is more of a personal preference and maybe I should be using jest --watch. I can revert if you like!

tests/infra/AutocompleteServiceImpl.ts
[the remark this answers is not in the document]
oshi97: If we remove the endpoint defaulting from within the services, we'll need to an Object.assign into either AnswersCore or provideCore for user defined endpoints and default endpoints. I still would want coverage for that, which we don't have any for now, but it would be pretty simple.
I thought about just checking the member variables on the services, but they're private. I don't think we want to make them public just for testing purposes.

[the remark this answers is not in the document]
oshi97: tldr; I like option 1 and making the services dumber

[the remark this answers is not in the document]
oshi97: Oh I didn't see your "Personally, I lean to 1." comment, thanks github

src/provideCore.ts
oshi97: this statement is no longer true

package.json
oshi97: I reverted

src/models/core/AnswersConfig.ts
```diff
@@ -77,3 +77,8 @@ export interface AnswersConfigWithToken extends BaseAnswersConfig {
  * @public
  */
 export type AnswersConfig = AnswersConfigWithApiKey | AnswersConfigWithToken;
+
+/**
+ * The AnswersConfig after any defaulting has been done from within answers-core.
+ */
+export type AnswersConfigWithDefaulting = AnswersConfig & { endpoints: Required<Endpoints> };
```
tmeyer2115, line 84: Could this be private/internal?

oshi97: I think the `@internal` annotations are only necessary for library exports that would otherwise be publicly exported, and this type isn't exported by the library. I was thinking it could be confusing to add it, since it could imply we're exporting it from the library but keeping it hidden, when in reality we're not. On the other hand it looks like we've already marked a few other things as internal, and leaving it unannotated also could be confused between intentionally leaving it unannotated and just forgetting to do so, so I'll add it!

tmeyer2115 (approved):

PR #172, opened 2022-06-17
added THIRD-PARTY-NOTICES to files in package.json

alextaing:
Added THIRD-PARTY-NOTICES to files in package.json.

J=SLAP-2070
TEST=manual

Tested using `npm pack` then `npm install` to install the .tgz file.  Checked that THIRD-PARTY-NOTICES was present.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "dist",
     "legacy",
     "lib",
-    "src"
+    "src",
+    "THIRD-PARTY-NOTICES"
   ],
   "engines": {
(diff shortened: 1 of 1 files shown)


oshi97 (approved):

oshi97 (changes requested):
actually for testing these can you use `npm pack` to create a tar file, and then run `npm install ./path-to-tar-file` in a test repo to make sure the third party notices show up?

oshi97 (approved):

PR #178, opened 2022-07-05
Rename from Answers to Search

alextaing:
WIP!  Answers => Search


src/SearchCore.ts
```diff
@@ -25,7 +25,7 @@ export class SearchCore {
   ) {}
 
   /**
-   * Performs an Answers search across all verticals.
+   * Performs an Search across all verticals.
```
oshi97, line 28: an -> a, also a little weird for search to be capitalized here

oshi97: oh this is still a WIP, if you open the PR as a draft from the start it won't automatically add reviewers

PR #181, opened 2022-07-05
Answers to Search (3): Rename request, error, and core

alextaing:
Rename the following interfaces:
AnswersRequest -> SearchRequest
AnswersError -> SearchError
AnswersCore -> SearchCore

### DIFF
--- a/src/models/deprecated.ts
+++ b/src/models/deprecated.ts
@@ -1,7 +1,14 @@
 import { SearchError } from './answersapi/SearchError';
+import { SearchRequest } from './core/SearchRequest';
 
 /**
  * @deprecated AnswersError is deprecated and has been replaced by SearchError
  * @public
  */
 export type AnswersError = SearchError;
+
+/**
+* @deprecated AnswersRequest is deprecated and has been replaced by SearchRequest
+* @public
+*/
+export type AnswersRequest = SearchRequest;
--- a/src/models/deprecated.ts
+++ b/src/models/deprecated.ts
@@ -1,0 +1,7 @@
+import { SearchError } from './answersapi/SearchError';
+
+/**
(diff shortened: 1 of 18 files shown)


src/models/deprecated.ts
```diff
@@ -0,0 +1,14 @@
+import { SearchError } from './answersapi/SearchError';
+import { SearchRequest } from './core/SearchRequest';
+
+/**
+ * @deprecated AnswersError is deprecated and has been replaced by SearchError
```
oshi97, line 5: `@link SearchError` which I think should work

src/models/deprecated.ts
```diff
@@ -0,0 +1,14 @@
...
+ * @public
+ */
+export type AnswersError = SearchError;
+
+/**
+* @deprecated AnswersRequest is deprecated and has been replaced by SearchRequest
+* @public
+*/
+export type AnswersRequest = SearchRequest;
```
nmanu1, line 14: using types for the synonyms could be a potentially breaking change. since `AnswersRequest` used to be an interface, defining it as a type here might break the implementation of anyone who previously used to extend `AnswersRequest`, because types can't be extended. I think it would be better to try to have the synonyms match if possible (i.e. interfaces have interface synonyms, classes have class synonyms, etc)

src/models/deprecated.ts
alextaing: Done!

src/models/deprecated.ts
alextaing: Done!

oshi97:
Could we update the Full Documentation link here to point to the new docs page? The current URL is no longer valid
image

oshi97 (approved):
I'd also update the commit msg to say you updated the readme but looks great!

PR #188, opened 2022-07-11
Make Result a generic

nmanu1:
Update `Result` to accept a generic type param that is passed to `rawData`.

J=SLAP-2231
TEST=auto, manual

Check that tests still pass and that the type param is optional. The default value is used unless a type is specified.

### DIFF
--- a/docs/search-core.result.entitytype.md
+++ b/docs/search-core.result.entitytype.md
@@ -4,7 +4,7 @@
 
 ## Result.entityType property
 
-The entity type of the result
+The entity type of the result.
 
 <b>Signature:</b>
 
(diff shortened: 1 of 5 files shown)


oshi97:
Do the parent response objects and/or search services also need to take in a generic?

nmanu1:
> Do the parent response objects and/or search services also need to take in a generic?

No, I made the the generic param optional by including a default, so the parent response objects aren't required to specify a type. Based on the item description, I assume we only want `Result` to accept a generic so that it can be used in the component library's `CardProps` to give compile-time checking when users write custom cards

oshi97 (approved):

PR #192, opened 2022-07-25
Add matchedSubstrings to spellCheck interface

alextaing:
[Raiden needs access](https://yext.slack.com/archives/C03NQ9X8ACB/p1658505872261039) to `matchedSubstrings` from the spellCheck part of the Search API response.  Since changing `correctedQuery` from a string to an object would be a breaking change, we added a new property to the spellCheck interface that holds the `matchedSubstrings`.  `matchedSubstrings` was also added to the Core response when the spellCheck object is created in `createSpellCheck`.

J=SLAP-2274
TEST=manual

It was verified that when spellCheck was provided in the Core response, that `matchedSubstrings` was also provided as its own property.


oshi97 (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-core",
-  "version": "1.8.1",
+  "version": "1.8.2",
```
oshi97, line 3: you should bump the version to a beta, like 1.9.0-#whateverPrNumberThisIs (I'm on github mobile right now)

This would be 1.9.0 not 1.8.2 because this is a feature not a bug fix

I was thinking you would bump it in a separate PR, but this also works!

oshi97 (changes requested):

yen-tt (commented):
to confirm, you want to publish a beta version right? This should point to develop instead of master

src/transformers/searchservice/createSpellCheck.ts
```diff
@@ -5,5 +5,6 @@ export function createSpellCheck(data: any): SpellCheck {
     originalQuery: data.originalQuery,
     correctedQuery: data.correctedQuery.value,
     type: data.type,
+    matchedSubstrings: data.correctedQuery.matchedSubstrings
```
yen-tt, line 8: can `data.correctedQuery.matchedSubstrings` be undefined?

package.json
alextaing: oke! bumped it back to 1.8.1 and we can bump version in a separate PR

src/transformers/searchservice/createSpellCheck.ts
alextaing: Looks like [product says](https://yext.slack.com/archives/C032CKFARGS/p1658842703108109) that a spell check object will always have `matchedSubstrings`!

yen-tt (approved):

PR #200, opened 2022-08-10
make value generic and create fieldType enum for FieldValueDirectAnswer

juliannzhou:
This PR makes the value of `DirectAnswer` generic and creates an `BuiltInFieldType` enum for use in fieldType for `FieldValueDirectAnswer` and `FeatureSnippedDirectAnswer`.

J=SLAP-2308
TEST=manual

Test: created a test interface with type `FieldValueDirectAnswer` and pass in `<number>` as the generic type variable, the interface works as expected when setting the value to a number with no errors.


src/models/searchservice/response/DirectAnswerFieldType.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Type for value of DirectAnswer.
+ *
+ * @public
+ */
+export enum DirectAnswerFieldType {
```
nmanu1, line 6: will this be used in `FieldValueDirectAnswer` for the generic type param?

juliannzhou: No, this will be used when we create specific direct answer types like when we create `UrlDirectAnswers`, `AddressDirectAnswers`.., we would use the corresponding field of `FieldValueDirectAnswer` as the `fieldType`

tmeyer2115, line 6: Could we call this simply `FieldType`? That's how it's referred to on the back-end. This is really a list of profile field types. We may use it for things besides direct answers in the future.

src/models/searchservice/response/FieldValueDirectAnswer.ts
```diff
@@ -7,11 +7,11 @@ import { Result } from './Result';
  *
  * @public
  */
-export interface FieldValueDirectAnswer extends DirectAnswer {
+export interface FieldValueDirectAnswer<T = unknown> extends DirectAnswer<T> {
```
tmeyer2115, line 10: Do we need the `T=unknown` here if we already have it in `DirectAnswer`?

juliannzhou: Yes, I tested it out and `T=unknown` is still needed else it errors out because we have to define the type variable in interface declaration to use T. But the `<T>` for DirectAnswer can be removed without errors.

src/models/searchservice/response/FieldType.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Type for value of DirectAnswer.
+ *
+ * @public
+ */
+export enum DirectAnswerFieldType {
```
tmeyer2115, line 6: Did the rename to `FieldType` get reverted?

tmeyer2115: Oh, I see. The file got renamed. I also wanted to rename the `enum`.

src/models/searchservice/response/DirectAnswer.ts
```diff
@@ -6,7 +6,7 @@ import { DirectAnswerType } from './DirectAnswerType';
  *
  * @public
  */
-export interface DirectAnswer {
+export interface DirectAnswer<T = unknown> {
```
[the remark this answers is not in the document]
juliannzhou: I think Yen wanted to leave the type as string for FieldType here as it is possible that we get other types, and then as we add specific Direct Answers (AddressDirectAnswers, PhoneDirectAnswers...) we would use the enum fields.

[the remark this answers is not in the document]
tmeyer2115: I think we do want to have `fieldType` use the new `FieldType` enum here. Davish let me know there are different types of featured snippet as well. So, I think we'd use `fieldType` there too to discriminate between the different types of snippet DA.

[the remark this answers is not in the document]
oshi97: I was wondering if we needed to worry about `FieldType` being a union type string vs an enum. long story short enum should be good! My primary concern was whether user's could switch on a type that is either enum or string so I wrote a short test
```ts
enum OnionTypes {
  Sweet = 'sweet',
  Red = 'Red',
  Negi = 'negi!!!!'
}

type Onion = OnionTypes | string;

function handle(o: Onion) {
  switch (o) {
    case 'sweet':
      console.log('sweet', o);
      break;
    case OnionTypes.Negi:
      console.log('negi', o);
      break;
    default:
      console.log('default', o);
  }
}

it('ogres are like onions', async () => {
  handle(OnionTypes.Sweet);
  handle(OnionTypes.Negi);
  handle('unknown onion');
});
```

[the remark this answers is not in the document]
oshi97: but yeah @juliannzhou we'll probably want fieldType to be `fieldType: FieldType | string`

src/models/searchservice/response/DirectAnswer.ts
```diff
@@ -15,7 +15,7 @@ export interface DirectAnswer {
    * @remarks
    * A value will not be present if the {@link DirectAnswer.fieldType} is 'rich_text'.
    */
-  value?: string,
+  value?: T,
```
oshi97, line 18: I know it was like this before, but why is value optional?

src/models/searchservice/response/DirectAnswer.ts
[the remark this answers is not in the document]
yen-tt: I was thinking we could use the same `EnumOrLiteral` concept we had in headless so it would accept FieldType in enum or string form, instead of doing `| string`?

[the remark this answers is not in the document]
oshi97: maybe it's a recent typescript change but it's possible we don't need EnumOrLiteral anymore? It looks like typescript automatically does something like EnumOrLiteral now? 
image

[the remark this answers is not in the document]
yen-tt: I'm not sure if I understand but it looks like if you type a string form of an enum value, it wouldn't work?
Screen Shot 2022-08-11 at 12 24 26 PM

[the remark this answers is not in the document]
oshi97: I looked through a couple releases and didn't see anything, do you think you could dig a little bit @yen-tt to see if we need EnumOrLiteral still?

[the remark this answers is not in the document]
oshi97: I looked through a couple releases and didn't see anything, but I can dig a little deeper to check if we still need EnumOrLiteral

src/models/searchservice/response/DirectAnswer.ts
tmeyer2115: IIRC, when we get a rich text field value, the value attribute is not set.

src/models/searchservice/response/DirectAnswer.ts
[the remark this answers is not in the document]
oshi97: that makes sense, I'm down to use `EnumOrLiteral` here!

src/models/searchservice/response/FieldType.ts
```diff
@@ -0,0 +1,22 @@
...
+  URL = 'url',
```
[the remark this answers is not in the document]
juliannzhou: I had to leave one fieldType instead of commenting all so that EnumOrLiteral would work. I left "address" uncommented.

src/models/searchservice/response/FeaturedSnippetDirectAnswer.ts
```diff
@@ -18,7 +20,7 @@ export interface FeaturedSnippetDirectAnswer extends DirectAnswer {
   /** {@inheritDoc DirectAnswer.verticalKey} */
   verticalKey: string,
   /** {@inheritDoc DirectAnswer.fieldType} */
-  fieldType: string,
+  fieldType: EnumOrLiteral<FieldType>,
```
oshi97, line 23: isn't FeaturedSnippetDirectAnswer always a string type?

src/models/searchservice/response/FieldValueDirectAnswer.ts
```diff
@@ -1,23 +1,25 @@
...
   /** {@inheritDoc DirectAnswer.value} */
-  value: string,
+  value: T,
   /** {@inheritDoc DirectAnswer.relatedResult} */
   relatedResult: Result,
   /** {@inheritDoc DirectAnswer.verticalKey} */
   verticalKey: string,
   /** {@inheritDoc DirectAnswer.fieldType} */
-  fieldType: string,
+  fieldType: EnumOrLiteral<FieldType>,
```
oshi97, line 22: Should fieldType here (and in the other places) be `fieldType: EnumOrLIteral<FieldType> | string` in order to handle custom field types, or future built in fieldTypes we don't have handling for?

oshi97: I was thinking we could even use a template literal type like
```ts
EnumOrLIteral<FieldType> | `c_${string}`
```
but it doesn't seem super useful and I don't really want to make assumptions about live API

src/models/searchservice/response/FieldValueDirectAnswer.ts
```diff
@@ -1,23 +1,25 @@
...
+import { FieldType } from './FieldType';
 import { Result } from './Result';
 
 /**
  * A direct answer where the answer came from a field from the knowledge graph.
  *
  * @public
  */
-export interface FieldValueDirectAnswer extends DirectAnswer {
+export interface FieldValueDirectAnswer<T = unknown> extends DirectAnswer<T> {
```
oshi97, line 12: how come `FieldValueDirectAnswer` extends `DirectAnswer<T>` with the generic passed through but `FeaturedSnippetDirectAnswer` doesn't need to pass the generic?

etc/search-core.api.md
```diff
@@ -109,11 +109,13 @@ export interface CombinedFilter {
 export type Context = any;
 
 // @public
-export interface DirectAnswer {
-    fieldType: string;
+export interface DirectAnswer<T = unknown> {
+    // Warning: (ae-forgotten-export) The symbol "EnumOrLiteral" needs to be exported by the entry point index.d.ts
```
nmanu1, line 113: do we want to export these to remove the errors?

src/models/searchservice/response/FeaturedSnippetDirectAnswer.ts
nmanu1, line 23: why does this field need to be specified again on `FeaturedSnippetDirectAnswer` when it's already in the extended `DirectAnswer` model? it doesn't look like we're changing the type either

src/models/searchservice/response/FieldType.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Type for value of DirectAnswer.
```
nmanu1, line 2: this doc seems incorrect. it should be for the `fieldType` of a `DirectAnswer`, right? also, can we add a link to `DirectAnswer`?

src/models/searchservice/response/FeaturedSnippetDirectAnswer.ts
nmanu1: looks like several fields in `FeaturedSnippetDirectAnswer` and `FieldValueDirectAnswer` are redefined without changes

src/models/searchservice/response/FieldValueDirectAnswer.ts
yen-tt: Decided to make a custom field with custom field type for Person entity in our test repo.
Screen Shot 2022-08-11 at 3 23 50 PM
Looks like it doesn't follow the format of c_blah, I think having EnumOrLiteral<FieldType> | string make sense. Although if we are doing | string, then having EnumOrLiteral is probably not needed.

src/models/searchservice/response/FeaturedSnippetDirectAnswer.ts
yen-tt: I believe someone mentioned in previous prs that our interfaces duplicate fields so it's more readable in the docs, although I'm not oppose to delete it now.

src/models/searchservice/response/FieldValueDirectAnswer.ts
yen-tt: Decided to make a custom field with custom field type for Person entity in our test repo.

Screen Shot 2022-08-11 at 3 30 02 PM

Looks like it doesn't follow the format of c_blah, I think having EnumOrLiteral<FieldType> | string make sense. Although if we are doing | string, then having EnumOrLiteral is probably not needed?
I think `FieldType | string` should be good!

oshi97: that sgtm! ... it's weird that fieldType is such an ugly string here

src/models/searchservice/response/FeaturedSnippetDirectAnswer.ts
oshi97: I also don't mind deleting the duplication. On one hand it's easier to read the code since you don't have to go up inheritance trees. On the other hand there's more code and it's more annoying to refactor stuff and write new code. I don't feel super strongly either way, readability of docs is nice but shouldn't have a large impact on how we write code. If we want better docs we can improve the way they are generated.

src/models/searchservice/response/FieldValueDirectAnswer.ts
yen-tt: I was thinking Julianne can probably wait for when the different types of `FeaturedSnippetDirectAnswer` should be address once product assign tasks to make changes for it then. But nonetheless, we can update this portion now

src/models/searchservice/response/FeaturedSnippetDirectAnswer.ts
yen-tt: asked tom, he's fine leaving it as is for now. If we decide to refactor later on, it wouldn't be a breaking change.

src/models/searchservice/response/FieldValueDirectAnswer.ts
yen-tt: updated!

oshi97: It wasn't a rhetorical question, I was a little unclear on what caused the differences. But if you think either way works lgtm!

oshi97 (approved):

PR #201, opened 2022-08-11
Rework static filters

nmanu1:
This PR reworks static filters to provide a better developer experience. A new `StaticFilters` model is added that replaces the existing `CombinedFilter`. `Filter` is renamed to `FieldValueFilter` in order to more accurately describe what the object represents. Now static filters can be either a field value filter or a nested object that is composed by combining field value filters. Also, I added Jest tests for some edge cases when serializing static filters.

J=SLAP-2325
TEST=auto, manual

See that the new and existing Jest tests pass. Spin up the test-site and see that requests can be successfully fired when specifying simple or nested static filters and the correct response is received.


src/models/searchservice/request/VerticalSearchRequest.ts
```diff
@@ -39,7 +38,7 @@ export interface VerticalSearchRequest extends SearchRequest {
   /** Enables session tracking. */
   sessionTrackingEnabled?: boolean,
   /** The static filters to apply to the search. */
-  staticFilters?: CombinedFilter | Filter,
+  staticFilters?: StaticFilter,
```
tmeyer2115, line 41: This feels a bit odd, the attribute name is plural, but we only allow a singular value.

src/models/searchservice/request/StaticFilter.ts
```diff
@@ -0,0 +1,22 @@
+import { FieldValueFilter } from './FieldValueFilter';
+
+/**
+ * Represents a static filter that will be used to refine results.
+ *
+ * @public
+ */
+export type StaticFilter =
+	{ kind: 'fieldValue', value: FieldValueFilter } |
+	{ kind: 'combination', combinator: FilterCombinator, children: StaticFilter[] };
```
tmeyer2115, line 10: Are these indentations correct?

src/models/searchservice/response/DisplayableFacet.ts
```diff
@@ -1,15 +1,11 @@
...
-import { NumberRangeValue } from '../common/NumberRangeValue';
 
 /**
  * A {@link Facet} which contains extra fields meant to be displayed to the end user.
  *
  * @public
  */
 export interface DisplayableFacet extends Facet {
-  /** {@inheritDoc Facet.fieldId} */
-  fieldId: string,
```
tmeyer2115, line 12: Is this now in `Facet`?

src/models/searchservice/response/DisplayableFacet.ts
```diff
@@ -22,10 +18,6 @@ export interface DisplayableFacet extends Facet {
  * @public
  */
 export interface DisplayableFacetOption extends FacetOption {
-  /** {@inheritDoc Matcher} */
-  matcher: Matcher,
-  /** {@inheritDoc FacetOption.value} */
-  value: string | number | boolean | NumberRangeValue,
```
tmeyer2115, line 28: These are now in `FacetOption`, right?

src/serializers/serializeStaticFilters.ts
```diff
@@ -1,32 +1,41 @@
...
-import { Filter } from '../models/searchservice/request/Filter';
+import { FilterCombinator, StaticFilter } from '../models/searchservice/request/StaticFilter';
+import { FieldValueFilter } from '../models/searchservice/request/FieldValueFilter';
 import { ApiStaticFilters, ApiFilter } from '../models/searchservice/request/ApiStaticFilters';
 
 export function serializeStaticFilters(
-  filter: CombinedFilter | Filter): string | undefined {
-  if (isCombinedFilter(filter)) {
-    return JSON.stringify(shapeCombinedFilterForApi(filter));
+  filter: StaticFilter): string | undefined {
```
tmeyer2115, line 7: Can we move this to the previous line?

src/serializers/serializeStaticFilters.ts
```diff
@@ -1,32 +1,41 @@
...
 
-  return JSON.stringify(shapeFilterForApi(filter));
+  return JSON.stringify(shapeFieldValueFilterForApi(filter.value));
 }
 
-function shapeCombinedFilterForApi(combinedFilter: CombinedFilter): ApiStaticFilters {
+function shapeCombinedFilterForApi(
+  combinator: FilterCombinator,
+  filters: StaticFilter[]
+): ApiStaticFilters | undefined {
```
tmeyer2115, line 19: Can the method declaration fit on one line?

src/models/searchservice/response/DisplayableFacet.ts
nmanu1: yes, this property was redundant even before the rework because it's the same as the property in `Facet`. there was no reason to specify it again after extending `Facet`

src/models/searchservice/response/DisplayableFacet.ts
nmanu1: this was the same thing as for `DisplayableFacet` where some properties were being redefined after extending the base model even though nothing was being changed. so I just removed the redundancy

src/serializers/serializeStaticFilters.ts
nmanu1: no, eslint complains about the line being too long if it's all on the same one

src/models/searchservice/request/StaticFilter.ts
nmanu1: updated!

src/serializers/serializeStaticFilters.ts
nmanu1: updated!

oshi97:
Are the main changes
1. renaming
2. a `kind` discriminator for filters, either `fieldValue` or `combination`?

nmanu1:
> Are the main changes
> 
> 1. a `kind` discriminator for filters, either `fieldValue` or `combination`
> 2. renaming associated with the above
> 
> ?

yes, the main things we wanted were
1. a single interface for static filters instead of having it explicitly split up into 2 types based on how we represent them for the API
2. the `kind` discriminator to make type checking easier for consumers of the Headless static filters state
3. renaming `Filter` to `FieldValueFilter` to make it more clear, because facets and static filters are filters too
4. and then the renaming associated with the above changes

src/models/searchservice/request/VerticalSearchRequest.ts
nmanu1: I agree, this looks a little weird. Although, this was the case before also where we accepted a single value (either a `Filter` or a `CombinedFilter`) for the plural `staticFilters`. It was just slightly obscured because of the union type.
Would it make sense to rename the attribute to `staticFilter`? Because that's what the user is giving us; it's just that they can combine a bunch of field value filters together if they want to in their one static filter

tmeyer2115: Yeah, renaming to static filter sounds good!

tmeyer2115 (approved):

PR #202, opened 2022-08-12
Custom fieldType for field value DA and builtin fieldType for feature snippet DA

yen-tt:
update direct answer's `fieldtype` to be: `fieldType: EnumOrLiteral<BuiltInFieldType> | 'unknown'`. This would support enum or literal string form of the KNOWN built in types and mark all the non built-in types as 'unknown'.

Also renamed `FieldValueDirectAnswer` to `BaseFieldValueDirectAnswer`. The new `FieldValueDirectAnswer` will be a union type of all built-in interfaces (e.g AddressFieldValueDA) and `CustomFieldValueDirectAnswer`

Move all direct answer related interfaces into its own folder `/directanswer` since there will be a decent amount for field value and potentially feature snippet ones in the future.

uncommented `phone` and `multi_line_text` in built-in types enum so the jest tests for createDirectAnswer for field value and feature snippet pass

Confirmed with product that FeaturedSnippetDirectAnswer can only have two field types: multi_line_text and rich_text. Updated interface to reflect this.

J=SLAP-2329
TEST=auto

added a new test for creating custom field value direct answer
add a new test to check for an error thrown if a feature snippet DA have a fieldType other than multi_line_text or rich_text


src/models/searchservice/response/directanswer/BaseFieldValueDirectAnswer.ts
```diff
@@ -1,14 +1,15 @@
...
+import { Result } from '../Result';
+import { EnumOrLiteral } from '../../../utils/EnumOrLiteral';
 
 /**
  * A direct answer where the answer came from a field from the knowledge graph.
  *
  * @public
  */
-export interface FieldValueDirectAnswer<T = unknown> extends DirectAnswer<T> {
+export interface BaseFieldValueDirectAnswer<T = unknown> extends DirectAnswer<T> {
```
oshi97, line 12: What do you think about putting FieldValueDirectAnswer and BaseFieldValueDirectAnswer into the FieldValueDirectAnswer file?

src/models/searchservice/response/directanswer/BaseFieldValueDirectAnswer.ts
```diff
@@ -1,14 +1,15 @@
...
   /** {@link DirectAnswerType}.FieldValue. */
   type: DirectAnswerType.FieldValue,
```
oshi97, line 14: I know we said previously we didn't care about removing duplicate properties, but I played around with it a bit more and I think it simplifies stuff a good bit so maybe we could do that here?

src/models/searchservice/response/directanswer/CustomFieldValueDA.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseFieldValueDirectAnswer } from './BaseFieldValueDirectAnswer';
+
+/**
+ * A {@link BaseFieldValueDirectAnswer} with a field type outside of {@link BuiltInFieldType}.
+ *
+ * @public
+ */
+export interface CustomFieldValueDA<T = unknown> extends BaseFieldValueDirectAnswer<T>{
```
yen-tt, line 8: not sure if we want to keep `DirectAnswer` or shortern to `DA` for the custom and built-in names. It might be inconsistent but the whole thing seems a bit long and I didn't want to cut off any part of the name.

src/models/searchservice/response/directanswer/BaseFieldValueDirectAnswer.ts
oshi97: I really like each file doing "one thing" but it's also nice for stuff to be more consolidated if we have like 10-20 line files

src/models/searchservice/response/directanswer/BaseFieldValueDirectAnswer.ts
yen-tt: Yea, it would certainly simplify things. I can remove duplicate for direct answer interfaces

src/models/searchservice/response/directanswer/BaseFieldValueDirectAnswer.ts
yen-tt: sure thing

tests/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -59,4 +61,42 @@ it('can create a FieldValueDirectAnswer', () => {
...
+      fieldApiName: 'phoneDescription',
+      fieldType: 'c010101.specialtype'
+    },
+    relatedItem: {
+      verticalConfigId: 'phone',
+      data: {}
+    }
+  };
+  const actualDirectAnswer = createDirectAnswer(apiFieldValueDirectAnswer);
+  const expectedDirectAnswer: CustomFieldValueDA = {
```
oshi97, line 85: I think we can sharpen the scope of this test if we use `expect.objectContaining` just to check that type is a DirectAnswerType.FieldValue and that `fieldType` is `'unknown'`

src/models/searchservice/response/directanswer/BaseFieldValueDirectAnswer.ts
yen-tt: hmm I kinda like them separate, but if you feel strongly, we can group them

tests/transformers/searchservice/createDirectAnswer.ts
yen-tt: good point!

tests/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -59,4 +61,42 @@ it('can create a FieldValueDirectAnswer', () => {
...
+    type: 'FIELD_VALUE',
+    answer: {
+      value: {
+        color: 'yellow',
+        os: 'android'
+      },
+      entityName: 'Obama Phone',
+      fieldName: 'Phone Description',
+      fieldApiName: 'phoneDescription',
+      fieldType: 'c010101.specialtype'
```
nmanu1, line 77: is it ok that we're replacing the value for a user's custom `fieldType` with 'unknown'? I'm wondering if it might be information that users want to be able to access somehow so they can differentiate between their custom fields

yen-tt: product is ok with it. User can access `fieldApiName` to check for their specific custom field and assert the typing from there!

nmanu1: oh ok, `fieldApiName` is another unique string that they define when creating a custom field?

src/models/searchservice/response/directanswer/BaseFieldValueDirectAnswer.ts
yen-tt: updated!

src/models/searchservice/response/directanswer/FeaturedSnippetDirectAnswer.ts
```diff
@@ -19,7 +20,7 @@ export interface FeaturedSnippetDirectAnswer<T = unknown> extends DirectAnswer<T
   /** {@inheritDoc DirectAnswer.verticalKey} */
   verticalKey: string,
   /** {@inheritDoc DirectAnswer.fieldType} */
-  fieldType: BuiltInFieldType | string,
+  fieldType: EnumOrLiteral<BuiltInFieldType> | 'unknown',
```
oshi97: Can FeaturedSnippetDirectAnswers also have different field types and values? I would have thought a snippet is always a string?

tests/transformers/searchservice/createDirectAnswer.ts
yen-tt: fieldApiName is the field name within an entity schema -- the fields you see generated when using the CLI command. User would provide a list of those field names (fieldApiName) to their experience to configure what can be use as a direct answer

tests/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -1,5 +1,7 @@
 import { createDirectAnswer } from '../../../src/transformers/searchservice/createDirectAnswer';
-import { DirectAnswerType } from '../../../src/models/searchservice/response/DirectAnswerType';
+import { DirectAnswerType } from '../../../src/models/searchservice/response/directanswer/DirectAnswerType';
+import { CustomFieldValueDA } from '../../../src/models/searchservice/response/directanswer/CustomFieldValueDA';
```
nmanu1, line 3: I thought the linting check was supposed to fail even on warnings like for this unused import?

Screen Shot 2022-08-12 at 2 44 21 PM

src/models/searchservice/response/directanswer/FeaturedSnippetDirectAnswer.ts
yen-tt: @tmeyer2115 talked to product, which they believe feature snippet can have other types

oshi97: Could we get a concrete example? I feel like we've tested FeaturedSnippetDirectAnswers before, but maybe not for non string types. It would be good for us to know how to do this in general, and if product doesn't know how to do this then that's probably a problem

tests/transformers/searchservice/createDirectAnswer.ts
yen-tt: looks like we have `--max-warnings=0` in the lint script command so it should fail I think. not sure why

oshi97: we don't have --max-warnings=0 set in the lint command here

src/models/searchservice/response/directanswer/FeaturedSnippetDirectAnswer.ts
oshi97: Arguably the fact that I don't know how to do it is a problem

src/models/searchservice/response/directanswer/BaseFieldValueDirectAnswer.ts
oshi97: sure separate is fine

tests/transformers/searchservice/createDirectAnswer.ts
yen-tt: my bad, looked at the wrong repo. I can add that in

tests/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -59,4 +59,26 @@ it('can create a FieldValueDirectAnswer', () => {
...
+    type: 'FIELD_VALUE',
+    answer: {
+      value: {
+        color: 'yellow',
+        os: 'android'
+      },
+      entityName: 'Obama Phone',
+      fieldName: 'Phone Description',
+      fieldApiName: 'phoneDescription',
+      fieldType: 'c010101.specialtype'
```
oshi97, line 75: this is a super nit but I don't think a businessId will ever start with a 0? (could be wrong) maybe we can use our test account's businessId 3350634

oshi97: I mean it's kind of exposing stuff in a public repo, but I don't really care too much about that personally...? some other semi-realistic business id also works for me

src/models/searchservice/response/directanswer/FeaturedSnippetDirectAnswer.ts
ElemelonWind: I think the rich text is a featured snippet. Here's an example
`"directAnswer":{
   "type":"FEATURED_SNIPPET",
   "answer":{
      "snippet":{
         "value":"# how to get rich text\n\ndrinks some coffee",
         "matchedSubstrings":[]
      },
      "fieldType":"rich_text"
   }
...`

ElemelonWind: that did not format the way i wanted but yeah it says type is FEATURED_SNIPPET

tests/transformers/searchservice/createDirectAnswer.ts
yen-tt: sure thing

src/models/searchservice/response/directanswer/CustomFieldValueDA.ts
oshi97: If I had to pick between the two I think I slightly prefer the long name. Ideally there's some way to rejigger the models so we don't have long names...? Having a hard time thinking of a solution though

yen-tt: I can update to the long name for now

src/models/searchservice/response/directanswer/CustomFieldValueDirectAnswer.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseFieldValueDirectAnswer } from './BaseFieldValueDirectAnswer';
+
+/**
+ * A {@link BaseFieldValueDirectAnswer} with a field type outside of {@link BuiltInFieldType}.
+ *
+ * @public
+ */
+export interface CustomFieldValueDirectAnswer<T = unknown> extends BaseFieldValueDirectAnswer<T>{
```
oshi97, line 8: Should this be `UnkownFieldValueDirectAnswer` instead of `CustomFieldValueDirectAnswer`? Since this isn't necessarily a custom field

src/models/searchservice/response/directanswer/FeaturedSnippetDirectAnswer.ts
```diff
@@ -12,14 +10,6 @@ import { Snippet } from './Snippet';
...
   /** {@link DirectAnswerType}.FeaturedSnippet. */
   type: DirectAnswerType.FeaturedSnippet,
-  /** {@inheritDoc DirectAnswer.value} */
-  value?: T,
-  /** {@inheritDoc DirectAnswer.relatedResult} */
-  relatedResult: Result,
-  /** {@inheritDoc DirectAnswer.verticalKey} */
-  verticalKey: string,
-  /** {@inheritDoc DirectAnswer.fieldType} */
-  fieldType: BuiltInFieldType | string,
```
oshi97, line 22: since davish told us the fieldType here can only be rich_text or multi_line_text should we reflect that here? I'm not sure whether we should also include unknown here but leaning towards not needing it?

yen-tt: yea, updating

src/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -11,7 +12,7 @@ export function createDirectAnswer(data: any): FeaturedSnippetDirectAnswer | Fie
     value: data.answer.value,
     relatedResult: ResultsFactory.fromDirectAnswer(data.relatedItem.data),
     verticalKey: data.relatedItem.verticalConfigId,
-    fieldType: data.answer.fieldType
+    fieldType: Object.values(BuiltInFieldType).includes(data.answer.fieldType) ? data.answer.fieldType : 'unknown'
```
oshi97, line 15: do we need to handle the case when data.answer.fieldType isn't `multi_line_text` or `rich_text` but `isFeaturedSnippetDirectAnswer` is true?

src/models/searchservice/response/directanswer/DirectAnswer.ts
```diff
@@ -14,13 +15,14 @@ export interface DirectAnswer<T = unknown> {
    * The result of the direct answer.
```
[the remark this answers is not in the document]
yen-tt: sure!

src/transformers/searchservice/createDirectAnswer.ts
yen-tt: hmm, we do expect from backend if it's a feature snippet then it should be multi_line_text or rich_text. We don't support unknown for feature snippet so I suppose we can throw an error

yen-tt: updated!

src/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -23,12 +24,16 @@ export function createDirectAnswer(data: any): FeaturedSnippetDirectAnswer | Fie
...
+      throw new Error(`Unexpected fieldType for featured snippet direct answer: ${fieldType}`);
+    }
     return {
       type: DirectAnswerType.FeaturedSnippet,
       ...commonDirectAnswerData,
       snippet: data.answer.snippet
     };
   } else {
-    throw new Error('The Search API returned an unknown direct answer type');
+    throw new Error(`The Search API returned an unknown direct answer type: ${data?.type}`);
```
oshi97, line 37: 💯

PR #203, opened 2022-08-12
create AddressDirectAnswer interface

juliannzhou:
This PR creates `AddressDirectAnswer` interface for direct answers with address values and an address type to define generic type variable.

J=SLAP-2310
TEST=manual

Test: created a test interface with `AddressDirectAnswer` and works as expected for when inputting address type values. The `Address` type interface is generated by CLI types command for Address field of Person entity. All address fields are verified by checking api response in Search except for `sublocality`, `region`, and `line 3`, as I cannot find a search result that returns an address direct answer with these three fields.


src/models/searchservice/response/AddressDirectAnswer.ts
```diff
@@ -0,0 +1,28 @@
+import { FieldValueDirectAnswer } from './FieldValueDirectAnswer';
+import { BuiltInFieldType } from './BuiltInFieldType';
+
+/**
+ * A interface for direct answers {@link FieldValueDirectAnswer} with address values.
```
nmanu1, line 5: nit: "An interface" instead of "A interface". also, because of the link text, this reads a little weirdly. maybe that part could be changed to something like `for a {@link FieldValueDirectAnswer} with`

src/models/searchservice/response/AddressDirectAnswer.ts
```diff
@@ -0,0 +1,28 @@
...
+ * An interface for direct answers for a {@link FieldValueDirectAnswer} with address values.
+ *
+ * @public
+ */
+export interface AddressDirectAnswer extends FieldValueDirectAnswer<Address> {
+  fieldType: BuiltInFieldType.Address
+}
+
+/**
+ * A type interface with address fields to use in {@link FieldValueDirectAnswer.value}.
```
oshi97, line 14: I think we can omit the calling this type interface

src/models/searchservice/response/AddressDirectAnswer.ts
```diff
@@ -0,0 +1,28 @@
+import { FieldValueDirectAnswer } from './FieldValueDirectAnswer';
+import { BuiltInFieldType } from './BuiltInFieldType';
+
+/**
+ * An interface for direct answers for a {@link FieldValueDirectAnswer} with address values.
```
nmanu1, line 5: this is a little wordy. I think it can be simplified by removing the first part ("An interface for direct answers for")

src/models/searchservice/response/AddressDirectAnswer.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface AddressDirectAnswer extends FieldValueDirectAnswer<Address> {
+  fieldType: BuiltInFieldType.Address
+}
+
+/**
+ * A type interface with address fields to use in {@link FieldValueDirectAnswer.value}.
+ *
+ * @public
+ */
+export interface Address {
```
oshi97, line 18: Out of curiosity are all of these properties truly optional?

juliannzhou: I tested it out on slanswers-test experience and I think line 1, city, region, and postalCode are required, although they are generated through CLI as optionals.

src/models/searchservice/response/AddressDirectAnswer.ts
```diff
@@ -0,0 +1,28 @@
+import { BuiltInFieldType } from './directanswer/BuiltInFieldType';
+import { DirectAnswer } from './directanswer/DirectAnswer';
+
+/**
+ * A direct answer {@link DirectAnswer} interface with address value.
```
nmanu1, line 5: can this be reworded? right now, the text would read "A direct answer DirectAnswer interface with ..."

oshi97:
what queries did you try? I wanted to try poking around as well

juliannzhou:
> @juliannzhou what queries did you try? I wanted to try poking around as well

I did 'yen address', 'bob address', 'event address', 'yext address'

src/models/searchservice/response/AddressDirectAnswer.ts
oshi97: What makes you think they're required?

juliannzhou: When you add address to a person or a location, these fields are required.

oshi97:
We need to add your new DirectAnswer interface to the FieldValueDirectAnswer union type. You can look at my pr as an example https://github.com/yext/search-core/pull/208 
If you want you can wait for my pr to be merged in, and then use the test-site infra I wrote

oshi97:
Could you add to the test-site's initDirectAnswer switch case?

src/models/searchservice/response/directanswer/fieldvaluedirectanswers/AddressDirectAnswer.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @public
+ */
+export interface Address {
+  line1: string,
+  line2?: string,
+  line3?: string,
+  sublocality?: string,
+  city: string,
+  region: string,
+  postalCode: string,
```
yen-tt, line 26: I'm not certain if all [non US areas](https://en.wikipedia.org/wiki/List_of_postal_codes) would have postal code? I added a location for Bolivia to the account test.

yen-tt: Screen Shot 2022-08-16 at 12 58 09 PM
Alright, looks like region and postalCode is not certain. let's make them optional.

oshi97: If we're not sure about things and composeAbstractSchema says they're all optional I don't mind just going with that

yen-tt: I would assume at least line1 should always be defined for an address? can't say the same for the other ones though.

yen-tt: Julianne talked to KM channel. Looks like countryCode should always be defined in practice and everything else is optional. So we should update to reflect that

src/models/searchservice/response/directanswer/fieldvaluedirectanswers/AddressDirectAnswer.ts
```diff
@@ -0,0 +1,29 @@
...
+export interface AddressDirectAnswer extends BaseFieldValueDirectAnswer<Address> {
+  fieldType: EnumOrLiteral<BuiltInFieldType.Address>
+}
+
+/**
+ * An interface with address fields to use in {@link BaseFieldValueDirectAnswer.value}.
+ *
+ * @public
+ */
+export interface Address {
```
yen-tt, line 19: you need to export this in the fieldvaluedirectanswer/index.ts file was well since this is part of the interface above

juliannzhou: I think it was added in the file. Can you check again if the file changes are refreshed?

yen-tt: I don't see `Address` in the index file. Also this warning indicates it's not exported.
Screen Shot 2022-08-16 at 3 26 32 PM

yen-tt: maybe you didn't push your new changes?

juliannzhou: Oh I see!

yen-tt (approved):

PR #204, opened 2022-08-15
Update static filters

nmanu1:
Update the `StaticFilter` interface to only allow combinations of filters supported by the backend. After talking with Watson, only `AND`s of `OR`s (conjunctions of disjunctions), or structures that can be trivially converted to such, are accepted. To reflect this, `ConjunctionStaticFilter` and `DisjunctionStaticFilter` models were created, with the latter having limitations on which filters it can combine so a disjunction of conjunctions is not allowed. Also, I added some Jest tests for a couple more edge cases that were being missed when serializing a static filter.

J=SLAP-2325
TEST=auto, manual

See that the new and existing Jest tests pass. Spin up the test-site and successfully fire vertical searches with static filters defined with the new interface.


src/models/searchservice/request/StaticFilter.ts
```diff
@@ -5,9 +5,49 @@ import { FieldValueFilter } from './FieldValueFilter';
...
+
+/**
+ * A {@link FieldValueFilter} with the kind of filter specified
+ * to discriminate between static filter types.
+ *
+ * @public
+ */
+export interface FieldValueStaticFilter extends FieldValueFilter {
+  /** The kind of static filter. */
+  kind: 'fieldValue'
```
oshi97, line 18: Have we thought about using an enum for filter kind? Not saying we should or shouldn't, I like the simplicity just making it a string is but we also seem to prefer enums in this library

nmanu1: I didn't really see a need because we don't currently refer to the `kind` values collectively. I think an enum makes sense in cases like for `Matcher` where we do refer to all the values as a single type that users can choose between

src/models/searchservice/request/StaticFilter.ts
```diff
@@ -5,9 +5,49 @@ import { FieldValueFilter } from './FieldValueFilter';
...
+/**
+ * A static filter composed by combining filters with the logical OR
+ * operator. The combined filters can either be field value filters or
+ * other disjunction filters.
+ *
+ * @public
+ */
+export interface DisjunctionStaticFilter {
+  /** {@inheritDoc FieldValueStaticFilter.kind} */
+  kind: 'disjunction',
```
oshi97, line 30: Do we need this kind when we have combinator?

src/models/searchservice/request/StaticFilter.ts
```diff
@@ -5,9 +5,49 @@ import { FieldValueFilter } from './FieldValueFilter';
...
+  /** {@inheritDoc FilterCombinator.OR} */
+  combinator: FilterCombinator.OR,
+  /** The filters to combine together. */
+  filters: (DisjunctionStaticFilter | FieldValueStaticFilter)[]
```
oshi97, line 34: So an $or filter is not able to have any $and's inside of it?

nmanu1: Exactly! That would be a disjunction of conjunctions, which is not allowed

oshi97: Wow I really thought they would allow for a single layer, but now I remember you said you switched a single $and and $or and it broke

oshi97 (approved):

PR #206, opened 2022-08-15
Add `TextValueDirectAnswer` interface.

tmeyer2115:
This PR creates the `TextValueDirectAnswer` interface, which describes a direct answer sourced from a string (or string list) field in the Knowledge Graph.

J=SLAP-2315
TEST=manual

I made a few code changes locally which are not included in this PR. These changes allowed me to generate a switch statement on `fieldType`. I added cases for `single_line_text` and `multi_line_text`. These cases were hit as expected when getting direct answers for string and string list fields. In these switch statements, I saw the IDE correctly inferring that the DA was a `TextValueDirectAnswer`.


oshi97:
Would you mind committing the test-site changes you talked about? I think it could be nice if every PR we added extra switch cases onto it

src/models/searchservice/response/directanswer/StringValueDirectAnswer.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseFieldValueDirectAnswer } from './BaseFieldValueDirectAnswer';
+import { BuiltInFieldType } from './BuiltInFieldType';
+
+/**
+ * A direct answer whose source is a string or string list field in the knowledge graph.
+ *
+ * @public
+ */
+export interface StringValueDirectAnswer extends BaseFieldValueDirectAnswer<string | string[]> {
+  fieldType: BuiltInFieldType.SingleLineText | BuiltInFieldType.MultiLineText
```
oshi97, line 10: Would single line text ever have a string[] value? For my edification what search queries did you use? (so I can store them in my brain for future use)

tmeyer2115: You can create a custom field that is single line text [], which translates to string[]. I just searched "Abington Services Offered". Services Offered was the single line text[] field.

oshi97, line 10: We may want to wrap fieldType in EnumOrLiteral? I'm not super happy with how wordy it is but 🤷

src/models/searchservice/response/directanswer/StringValueDirectAnswer.ts
```diff
@@ -0,0 +1,12 @@
+import { EnumOrLiteral } from '../../../utils/EnumOrLiteral';
+import { BaseFieldValueDirectAnswer } from './BaseFieldValueDirectAnswer';
+import { BuiltInFieldType } from './BuiltInFieldType';
+
+/**
+ * A direct answer whose source is a string or string list field in the knowledge graph.
+ *
+ * @public
+ */
+export interface StringValueDirectAnswer extends BaseFieldValueDirectAnswer<string | string[]> {
```
oshi97, line 10: I was thinking a number of field types (instagram handle, facebook url, linkedin url, etc.) are better suited for the name StringValueDirectAnswer, since their values can truly only be string and never string[] (afaict anyways).
Is there some other name we could give for this interface? I had a hard time thinking of one besides say `TextValueDirectAnswer` which didn't seem ideal for me.

tmeyer2115: We could also have an interface for those types and lock it down. For example, we'd have:
`FacebookUrlDirectAnswer extends StringValueDirectAnswer<string>`.

oshi97 (approved):

PR #207, opened 2022-08-15
add featured snippet DA interfaces for the builtin fieldType 

yen-tt:
- add `MultiLineTextFeaturedSnippetDirectAnswer` and `RichTextFeaturedSnippetDirectAnswer` for the two types that fieldType of featured snippet direct answer can be.
  - `multi_line_text` for feature snippet DA have non optional `value` field of type string
  - `rich_text` for feature snippet DA does not have `value` field so that is omitted in the interface
- the old `FeaturedSnippetDirectAnswer` is now the base interface `BaseFeaturedSnippetDirectAnswer` that the specific interfaces extend from. The new `FeaturedSnippetDirectAnswer` is the union type.

J=SLAP-2330
TEST=auto&manual

oshi confirmed the structure of multi_line_text and rich_text through examples listed in the [comments under this item](https://yexttest.atlassian.net/browse/SLAP-2317) where rich_text does NOT provide `value` field and `multi_line_text` does.

tested locally that:
- 'value' field is not present in IDE for the switch case 'richtext'
Screen Shot 2022-08-15 at 1 19 51 PM
- 'value' field is present, of type string, and NOT optional in IDE for the switch case 'multi_line_text'
Screen Shot 2022-08-15 at 1 21 57 PM


src/models/searchservice/response/directanswer/RichTextFeaturedSnippetDirectAnswer.ts
```diff
@@ -0,0 +1,14 @@
...
+import { BaseFeaturedSnippetDirectAnswer } from './BaseFeaturedSnippetDirectAnswer';
+import { EnumOrLiteral } from '../../../utils/EnumOrLiteral';
+
+/**
+ * A {@link BaseFeaturedSnippetDirectAnswer} with 'rich_text' field type.
+ * "value" field is omitted for featured snippet direct answer of this field type.
+ *
+ * @public
+ */
+export interface RichTextFeaturedSnippetDirectAnswer extends Omit<BaseFeaturedSnippetDirectAnswer<string>, 'value'> {
```
oshi97, line 11: are you sure you don't want to put RichTextFeaturedSnippetDirectAnswer, MultiLineTextFeaturedSnippetDirectAnswer, and BaseFeaturedSnippetDirectAnswer into FeaturedSnippetDirectAnswer.ts ? There's a lot of jumping around when I try to navigate the code, and each file is less than 20 lines and I don't see them getting much bigger in the future

src/models/searchservice/response/directanswer/FeaturedSnippetDirectAnswer.ts
```diff
@@ -1,18 +1,9 @@
-import { DirectAnswer } from './DirectAnswer';
-import { DirectAnswerType } from './DirectAnswerType';
-import { Snippet } from '../Snippet';
-import { BuiltInFieldType } from './BuiltInFieldType';
-
+import { MultiLineTextFeaturedSnippetDirectAnswer } from './MultiLineTextFeaturedSnippetDirectAnswer';
+import { RichTextFeaturedSnippetDirectAnswer } from './RichTextFeaturedSnippetDirectAnswer';
 /**
- * A direct answer which was found within a document.
+ * Possible built-in {@link BaseFeaturedSnippetDirectAnswer} interfaces
```
oshi97, line 4: Can we change "Possible" to "All possible" (or something similar to show completeness)?

src/models/searchservice/response/directanswer/RichTextFeaturedSnippetDirectAnswer.ts
oshi97: It would only add about ~30 lines to FeaturedSnippetDirectAnswer.ts, I feel like that's totally reasonable

src/models/searchservice/response/directanswer/MultiLineTextFeaturedSnippetDirectAnswer.ts
```diff
@@ -0,0 +1,15 @@
+import { BuiltInFieldType } from './BuiltInFieldType';
+import { BaseFeaturedSnippetDirectAnswer } from './BaseFeaturedSnippetDirectAnswer';
+import { EnumOrLiteral } from '../../../utils/EnumOrLiteral';
+
+/**
+ * A {@link BaseFeaturedSnippetDirectAnswer} with 'multi_line_text' field type
```
nmanu1, line 6: nit: missing period

src/models/searchservice/response/directanswer/MultiLineTextFeaturedSnippetDirectAnswer.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @public
+ */
+export interface MultiLineTextFeaturedSnippetDirectAnswer extends BaseFeaturedSnippetDirectAnswer<string> {
+  /** The value of the direct answer. */
```
nmanu1, line 11: can this doc be inherited from the base direct answer interface?

src/models/searchservice/response/directanswer/BaseFeaturedSnippetDirectAnswer.ts
```diff
@@ -0,0 +1,19 @@
...
+import { BuiltInFieldType } from './BuiltInFieldType';
+import { EnumOrLiteral } from '../../../utils/EnumOrLiteral';
+
+/**
+ * A direct answer which was found within a document.
+ *
+ * @public
+ */
+export interface BaseFeaturedSnippetDirectAnswer<T = unknown> extends DirectAnswer<T> {
+  /** {@link DirectAnswerType}.FeaturedSnippet */
```
nmanu1, line 13: should this be an `inheritDoc` instead of a `link`? as it is right now, this doc doesn't seem to be adding much value

src/models/searchservice/response/directanswer/RichTextFeaturedSnippetDirectAnswer.ts
yen-tt: sure, I can combine them

src/models/searchservice/response/directanswer/MultiLineTextFeaturedSnippetDirectAnswer.ts
yen-tt: There's a remark in the base direct answer interface related to 'rich_text', that I didn't want since it's not applicable to multi line text field type, so I opted to have the doc here directly instead of inheriting it.

src/models/searchservice/response/directanswer/BaseFeaturedSnippetDirectAnswer.ts
yen-tt: this was included before this pr. I don't think the dot notation here works anyway. inheritDoc sounds good. I can update

oshi97 (approved):
lgtm!

PR #209, opened 2022-08-16
create phone direct answer

juliannzhou:
This PR creates PhoneFieldValueDirectAnswer interface for direct answers with phone values.

J=SLAP-2316
TEST=manual

Test: created a test interface using PhoneFieldValueDirectAnswer. When inputting phone type values, the type works as expected. The phone type is verified by api response in Search.


test-site/src/ts/initDirectAnswers.ts
```diff
@@ -49,6 +49,9 @@ const FIELD_VALUE_QUERIES = {
   ],
   [BuiltInFieldType.Integer]: [
     'oliver shi outdoor pools?' // number
+  ],
+  [BuiltInFieldType.Phone]: [
+    'oliver shi phone?' // string
```
oshi97, line 54: can we add "amani farooque phone number" here? it's kind of a classic

juliannzhou: okay i'll add it XD

oshi97 (approved):

oshi97:
nice!

PR #213, opened 2022-08-16
fix compilation_check_for_test_site

oshi97:
In a github workflow the `cd` command only affects the current step. This was causing false positives.
Also fix a casing issue with a test-site import on the js side.

J=NONE
TEST=manual

checked out a branch that should be failing, saw it fail locally, but succeed in the PR it's in
in the github workflow logs for the PR it's in, I see the regular search-core build being run twice,
and the test-site build not being run


.github/workflows/run-tests.yml
```diff
@@ -19,6 +19,4 @@ jobs:
         uses: actions/setup-node@v3
       - run: npm ci
       - run: npm run build
-      - run: cd test-site
-      - run: npm i
-      - run: npm run build
+      - run: cd test-site && npm i && npm run build
```
nmanu1, line 22: could this job use the reusable run-tests workflow and pass this last step in as the test script?

.github/workflows/run-tests.yml
```diff
@@ -12,13 +12,7 @@ jobs:
...
-      - name: Use Node.js
-        uses: actions/setup-node@v3
-      - run: npm ci
-      - run: npm run build
-      - run: cd test-site
-      - run: npm i
-      - run: npm run build
+    uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
+    with:
+      test_script: cd test-site && npm i && npm run build
```
nmanu1, line 17: would setting the test-site as the working directory variable for the reusable workflow work? or do we need to do a fresh install and build of the top-level first?

oshi97: Yeah we need to build the search-core assets in the base folder first.

nmanu1 (approved):

PR #217, opened 2022-08-17
remove alpha from additionalQueryParameter and make public

juliannzhou:
This PR removes @alpha tag from `additionalQueryParameter` and makes it public.

J=SLAP-1398
TEST=manual

Original PR (https://github.com/yext/search-core/pull/79) made this property alpha to prevent access from typescript users of core, but this property can be public after release.


oshi97 (approved):

src/models/core/SearchConfig.ts
```diff
@@ -31,7 +31,7 @@ export interface BaseSearchConfig {
   /**
    * Additional query params added on to every request.
    *
-   * @alpha
+   * @public
```
nmanu1, line 34: can we just remove this tag now since it matches the parent interface's release tag?

PR #222, opened 2022-09-06
Version 2.0

yen-tt:


oshi97:
I think all the deprecated code that was removed was also removed for rebranding purposes, so we could make a note of that? Otherwise somebody might be wondering if there was other stuff that we had deprecated

yen-tt:
> I think all the deprecated code that was removed was also removed for rebranding purposes, so we could make a note of that? Otherwise somebody might be wondering if there was other stuff that we had deprecated. If the list isn't too long I think it would be good to have them explicitly written out here as well

sure! Updated PR description to note that the deprecated exports removed was part of the rebranding process. Instead of listing all of the deprecated identifiers here, I linked to the [V1.8 release notes](https://github.com/yext/search-core/releases/tag/v1.8.0) that have a list of all those stuff

oshi97 (approved):

PR #230, opened 2023-03-23
Add inputIntents to AutocompleteResult

oshi97:
Adds AutocompleteResult.inputIntents.

T=TECHOPS-8462
TEST=auto,manual

saw that this would fix the related techops for missing query intents


src/models/autocompleteservice/AutocompleteResponse.ts
```diff
@@ -93,5 +93,9 @@ export interface AutocompleteResult {
    * Any vertical keys associated with a prompt.
    * This only shows up on universal autocomplete requests.
    **/
-  verticalKeys?: string[]
+  verticalKeys?: string[],
+  /**
+   * {@link SearchIntent}s corresponding to the autocomplete result.
+   */
+  queryIntents: SearchIntent[]
```
nmanu1, line 100: What's the reason for naming this field `queryIntents`, when the similar field in `AutocompleteResponse` is called `inputIntents`?

tmeyer2115 (approved):

PR #233, opened 2023-05-03
Provide Endpoints for EU Cloud Region

EmilyZhang777:
This PR creates an EndpointProvider class to offer a way to get endpoints for EU.

J=BACK-2270
TEST=auto,manual

Ran tests under tests/.
Manually started a test site and made sure the api request use the correct endpoints.

Change-Id: I5c5bbd99e9c3f0f67d0b6817e0a0db6eeaf2f231


src/constants.ts
```diff
@@ -2,27 +2,125 @@ import { Endpoints } from './models/core/Endpoints';
...
+
+  constructor(environment?: Environment, cloudRegion?: CloudRegion, overrideEndpoints?: Endpoints) {
+    this.environment = environment || Environment.PROD;
+    this.cloudRegion = cloudRegion || CloudRegion.US;
+    this.overrideEndpoints = overrideEndpoints;
+  }
+
+  /** @internal */
+  getDomain() {
+    return `https://${this.environment}-cdn.${this.cloudRegion}.yextapis.com`;
```
benmcginnis, line 33: which domain is this? i don't think it is the right one for us live API?

src/constants.ts
```diff
@@ -2,27 +2,125 @@ import { Endpoints } from './models/core/Endpoints';
...
+      `${this.getDomain()}/v2/accounts/me/search/filtersearch`;
+  }
+}
+
+/**
+ * The endpoints to use for sandbox experiences.
+ * @deprecated Use {@link EndpointsProvider} instead
+ * @public
+ */
+export const SandboxEndpoints: Required<Endpoints> =
```
benmcginnis, line 125: i think we also need to export the default endpoints since that could be used by someone?

src/provideCore.ts
```diff
@@ -24,10 +24,8 @@ export function provideCore(config: SearchConfig): SearchCore {
 
   const defaultedConfig: SearchConfigWithDefaulting = {
     ...config,
-    endpoints: {
-      ...defaultEndpoints,
-      ...config.endpoints
-    }
+    endpoints:
+      new EndpointsProvider(Environment.PROD, CloudRegion.US, config.endpoints).getEndpoints()
```
benmcginnis, line 28: taking the config object would simplify this

tests/infra/QuestionSubmissionServiceImpl.ts
```diff
@@ -42,7 +42,7 @@ describe('Question submission', () => {
   });
 
   it('uses the production endpoint by default', () => {
-    const expectedUrl = 'https://liveapi.yext.com/v2/accounts/me/createQuestion';
+    const expectedUrl = 'https://prod-cdn.us.yextapis.com/v2/accounts/me/createQuestion';
```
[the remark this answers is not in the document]
EmilyZhang777: yes, liveapi.yext.com will still be supported though. [Here's](https://docs.google.com/document/d/1pu4IhzubaChPPMX5Vmu3LeXjOfyIR6cWBx2YfNizi_E/edit#) the new naming convention.

src/constants.ts
```diff
@@ -2,27 +2,125 @@ import { Endpoints } from './models/core/Endpoints';
...
-} as const;
+export enum CloudRegion {
+  US = 'us',
+  EU = 'eu',
+}
+
+export enum Environment {
+  PROD = 'prod',
+  SANDBOX = 'sbx',
+}
```
cea2aj, line 13: can you add some docs around these? They should be publicly exported so there will be documentation generated for them

src/constants.ts
```diff
@@ -2,27 +2,125 @@ import { Endpoints } from './models/core/Endpoints';
...
-  verticalAutocomplete: 'https://liveapi-sandbox.yext.com/v2/accounts/me/search/vertical/autocomplete',
-  filterSearch: 'https://liveapi-sandbox.yext.com/v2/accounts/me/search/filtersearch',
-} as const;
\ No newline at end of file
+export class EndpointsProvider {
+  private readonly environment: Environment;
+  private readonly cloudRegion: CloudRegion;
+  private readonly overrideEndpoints?: Endpoints;
+
+  constructor(environment?: Environment, cloudRegion?: CloudRegion, overrideEndpoints?: Endpoints) {
```
cea2aj, line 25: I'm not sure that overrideEndpoints is necessary here. Someone could manually set additional override outside of this class. for example: 
```
const providedEndpoints = new EndpointsProvider(
      Environment.SANDBOX,
      CloudRegion.US
).getEndpoints();

const customEndpoints = {
   ...proviedEndpoints,
   verticalSearch: '<verticalSearchEndpoint>'
}
```

src/constants.ts
EmilyZhang777: defaultEndpoints wasn't exposed through index so I assumed it's not used publicly. I've updated all internal usages. Let me know what you think!

etc/search-core.api.md
```diff
@@ -263,6 +263,31 @@ export interface Endpoints {
     verticalSearch?: string;
 }
 
+// @public
+export class EndpointsProvider {
+    // Warning: (ae-forgotten-export) The symbol "Environment" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-forgotten-export) The symbol "CloudRegion" needs to be exported by the entry point index.d.ts
+    constructor(environment?: Environment, cloudRegion?: CloudRegion, overrideEndpoints?: Endpoints);
```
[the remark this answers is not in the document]
cea2aj: the risk with requiring the config object is that this class would depend on many required properties which aren't relevant to this class. an alternative would be have the constructor take in a single object with environment and cloudRegion properties. that way if we were to update the config to have environment and cloudRegion properties, the config could be passed in and typescript would allow it based on duck typing

src/constants.ts
```diff
@@ -2,27 +2,125 @@ import { Endpoints } from './models/core/Endpoints';
...
+      universalAutocomplete: this.getUniversalAutocomplete(),
+      verticalAutocomplete: this.getVerticalAutocomplete(),
+      filterSearch: this.getFilterSearchEndpoint(),
+    };
+  }
+
+  /**
+   * Returns the endpoint for universal search.
+   * @public
+   * */
```
cea2aj, line 60: Is it necessary to make all of these public as well? I think just making `getEndpoints` public is sufficient. Is there a use case where someone would be calling these individual methods?

src/constants.ts
cea2aj: I don't think we need to export the default endpoints because they will be used if no custom endpoints are specified

tests/infra/SearchServiceImpl.ts
```diff
@@ -157,10 +160,13 @@ describe('SearchService', () => {
       const customUrl = 'http://custom.endpoint.com/api';
       const config: SearchConfigWithDefaulting = {
         ...configWithRequiredApiKey,
-        endpoints: {
-          ...defaultEndpoints,
-          universalSearch: customUrl
-        }
+        endpoints:
+          new EndpointsProvider(
```
cea2aj, line 164: rather than making the EndpointsProvider public, I think it's worth considering making it internal and instead exporting a `provideEndpoints` function which calls the EndpointsProvider internally. That would follow the existing naming convention in our repos  e.g. (`provideCore` and `provideHeadless`), and it would be simpler since they wouldn't need to construct a class. For example, the call would look like `provideEndpoints(Environment.PROD, CloudRegion.US`.

etc/search-core.api.md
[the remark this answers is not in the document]
EmilyZhang777: I agree with Connor. I'll remove the overrideEndpoints and callers can implement their own override if that sounds good to people.

src/constants.ts
EmilyZhang777: Added a public provideEndpoints function per Connor's comment

src/constants.ts
EmilyZhang777: It's the new naming convention: https://docs.google.com/document/d/1pu4IhzubaChPPMX5Vmu3LeXjOfyIR6cWBx2YfNizi_E/edit#

src/provideCore.ts
EmilyZhang777: No longer relevant. Ack

src/constants.ts
EmilyZhang777: Done

tests/infra/SearchServiceImpl.ts
EmilyZhang777: Done

src/index.ts
```diff
@@ -1,6 +1,6 @@
 // main symbols
 export { SearchCore } from './SearchCore';
 export { provideCore } from './provideCore';
-export { SandboxEndpoints } from './constants';
+export { provideEndpoints, SandboxEndpoints } from './constants';
```
cea2aj, line 4: Can you export Environment and CloudRegion from here as well?

etc/search-core.api.md
```diff
@@ -263,6 +263,31 @@ export interface Endpoints {
     verticalSearch?: string;
 }
 
+// @public
+export class EndpointsProvider {
+    // Warning: (ae-forgotten-export) The symbol "Environment" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-forgotten-export) The symbol "CloudRegion" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
cea2aj: It looks like this warning is still here

src/constants.ts
```diff
@@ -2,27 +2,88 @@ import { Endpoints } from './models/core/Endpoints';
...
+  PROD = 'prod',
+  SANDBOX = 'sbx',
+}
+
+/**
+ * Provides methods for getting various endpoints.
+ *
+ * @internal
+ */
+export class EndpointsProvider {
```
cea2aj, line 30: nit: I think this would fit better outside of a constants file. Maybe we could make a `provideEndpoints` file, and move this class and that function there as well? I think `provideEndpoints.ts` would fit in either src or src/utils

cea2aj:
This is looking really good!

src/constants.ts
EmilyZhang777: Renamed the file 👍

EmilyZhang777: Renamed the file 👍

src/index.ts
EmilyZhang777: Ah, thought I did it. Now exported

cea2aj:
Since this repo uses git flow, this change should be based on develop, and the PR should target develop. When we are ready to release the next version, we bump the version number in develop in another PR and then merge develop into master

cea2aj (approved):
LGTM! Note: I don't think the change-id in the description is necessary. Also, when we merge into develop, we always use the squash and merge option so that these changes appear in develop in a single commit.

etc/search-core.api.md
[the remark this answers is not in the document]
benmcginnis: My suggestion then would be to move the relevant endpoint, region, and environment properties to an interface that the config object extends.  That way different sub components don't have to know about the whole config object, but users only need to know about one and can just pass it around.

src/provideEndpoints.ts
```diff
@@ -0,0 +1,89 @@
...
+  PROD = 'prod',
+  SANDBOX = 'sbx',
+}
+
+/**
+ * Provides methods for getting various endpoints.
+ *
+ * @internal
+ */
+export class EndpointsProvider {
```
benmcginnis, line 30: sorry my only other thought is that we might want to not call this a Provider since it is not a React Context Provider

EmilyZhang777: Done

src/provideEndpoints.ts
```diff
@@ -0,0 +1,69 @@
...
+import { ServingConfig } from './models/core/SearchConfig';
+
+export const defaultApiVersion = 20220511;
+
+/**
+ * Provides methods for getting various endpoints.
+ *
+ * @internal
+ */
+export class EndpointsAdapter {
```
benmcginnis, line 13: ok I'm sorry about the name again but since we're not sending the endpoints param into this class can we make it an EndpointsFactory instead?

benmcginnis (approved):

src/provideEndpoints.ts
EmilyZhang777: Of course. Updated!

etc/search-core.api.md
```diff
@@ -642,6 +645,20 @@ export interface SearchService {
...
+export interface ServingConfig {
+    // Warning: (ae-forgotten-export) The symbol "CloudRegion" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-unresolved-inheritdoc-reference) The @inheritDoc reference could not be resolved: The package "@yext/search-core" does not have an export "CloudRegion"
+    //
+    // (undocumented)
+    cloudRegion?: CloudRegion;
+    // Warning: (ae-forgotten-export) The symbol "Environment" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-unresolved-inheritdoc-reference) The @inheritDoc reference could not be resolved: The package "@yext/search-core" does not have an export "Environment"
+    //
+    // (undocumented)
```
cea2aj, line 658: It looks like CloudRegion and Environment need to be exported again

src/provideEndpoints.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Provides all endpoints based on environment and cloud region.
+ *
+ * @remarks
+ * Returns an {@link Endpoints} instance.
+ *
+ * @param config - serving config to use for the domain, defaults to Prod and US if not provided.
+ *
+ * @public
+ */
+export function provideEndpoints(config?: ServingConfig): Required<Endpoints> {
```
cea2aj, line 67: Since the environment and cloud region are part of the SearchConfig, I don't think we need this helper function anymore

src/provideEndpoints.ts
```diff
@@ -0,0 +1,69 @@
...
+  private readonly cloudRegion: CloudRegion;
+
+  constructor(config?: ServingConfig) {
+    this.environment = config?.environment || Environment.PROD;
+    this.cloudRegion = config?.cloudRegion || CloudRegion.US;
+  }
+
+  /** Provides the domain based on environment and cloud region.
+   * @internal
+   */
```
cea2aj, line 24: nit: I don't believe the internal tags are necessary on the methods because the class is marked as internal, so it is implied that the methods are as well

benmcginnis:
LGTM! I think we're good to squash & merge

src/models/core/SearchConfig.ts
```diff
@@ -1,11 +1,33 @@
...
+   */
+  cloudRegion?: CloudRegion
+}
 
 /**
  * The base configuration options for {@link SearchCore}.
  * @public
  */
-export interface BaseSearchConfig {
+export interface BaseSearchConfig extends ServingConfig {
```
cea2aj, line 30: even though BaseSearchConfig extends ServingConfig, it may be helpful to include the environment and cloudRegion properties in the BaseSearchConfig interface so that they appear in the generated docs.

cea2aj (changes requested):

src/provideEndpoints.ts
EmilyZhang777: Don't we still need this if we don't want to expose EndpointsFactory class?

src/models/core/SearchConfig.ts
benmcginnis: I think maybe if we `@link ServingConfig` in the json doc or do something similar it will pull the docs in if ServingConfig is exported

src/provideEndpoints.ts
EmilyZhang777: Oh I understand what you mean. So they will always use provideCore. Sure. I'll remove this

src/models/core/SearchConfig.ts
cea2aj: The generated docs link to the ServingConfig it extends it, but it doesn't include its properties on the BaseConfig interface doc page. In the past on Slapshot we made the decision to duplicate the properties for the sake of clearer docs

src/provideEndpoints.ts
EmilyZhang777: Done

etc/search-core.api.md
EmilyZhang777: Done

src/models/core/SearchConfig.ts
EmilyZhang777: Updated

src/provideEndpoints.ts
cea2aj: It looks like the function still exists?

EmilyZhang777: oops, forgot to include the file. Now actually removed!

PR #235, opened 2023-05-04
Remove Reference to provideCore in Doc

EmilyZhang777:
This PR removes the reference to provideCore in the doc since it is not re-exported in headless and was causing an issue.
It also bumps the beta version to 2.3.0-beta.1.

J=BACK-2270
TEST=auto

Ran tests under tests/.


cea2aj:
remember to target develop

cea2aj (approved):

PR #241, opened 2023-06-14
RichTextV2 and Markdown changes

saahithjanapati:
Added support for direct answers to return raw HTML of converted RichTextv2 and markdown in snippets

J=BACK-2241

TEST=manual

Tested using a modified version of the test site in the search-core repository. Also verified that the value field in direct answer snippet contained converted HTML for queries.


docs/search-core.basefeaturedsnippetdirectanswer.fieldtype.md
```diff
@@ -9,5 +9,5 @@ The field type of the direct answer.
 <b>Signature:</b>
 
 ```typescript
-fieldType: EnumOrLiteral<BuiltInFieldType.MultiLineText | BuiltInFieldType.RichText>;
+fieldType: EnumOrLiteral<BuiltInFieldType.MultiLineText | BuiltInFieldType.RichText | BuiltInFieldType.RichText_v2 | BuiltInFieldType.Html | BuiltInFieldType.Markdown>;
```
benmcginnis, line 12: Is HTML actually a built in field type?

src/models/searchservice/response/directanswer/BuiltInFieldType.ts
```diff
@@ -19,5 +19,8 @@ export enum BuiltInFieldType {
   Integer = 'integer',
   SingleLineText = 'single_line_text',
   RichText = 'rich_text',
-  MultiLineText = 'multi_line_text'
+  MultiLineText = 'multi_line_text',
+  RichText_v2 = 'rich_text_v2',
+  Html = 'html',
```
benmcginnis, line 24: I don't think html is actually a field type, is it?

src/models/searchservice/response/directanswer/FeaturedSnippetDirectAnswer.ts
```diff
@@ -5,7 +5,10 @@ import { BuiltInFieldType } from './BuiltInFieldType';
 import { EnumOrLiteral } from '../../../utils/EnumOrLiteral';
 import {
   MultiLineTextSnippetDirectAnswer,
-  RichTextSnippetDirectAnswer
+  RichTextSnippetDirectAnswer,
```
benmcginnis, line 8: I think they are all converted to html by the server response

docs/search-core.basefeaturedsnippetdirectanswer.fieldtype.md
cea2aj: yep!

src/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -25,7 +25,11 @@ export function createDirectAnswer(data: any): FeaturedSnippetDirectAnswer | Fie
     };
   } else if (isFeaturedSnippetDirectAnswer) {
     const fieldType = commonDirectAnswerData.fieldType;
-    if (fieldType != BuiltInFieldType.MultiLineText && fieldType != BuiltInFieldType.RichText) {
+    if (fieldType != BuiltInFieldType.MultiLineText &&
+        fieldType != BuiltInFieldType.RichText &&
+        fieldType != BuiltInFieldType.RichText_v2 &&
+        fieldType != BuiltInFieldType.Html &&
+        fieldType != BuiltInFieldType.Markdown) {
```
cea2aj, line 32: nit: can you put these built in field types in a list and call the `includes` method instead of checking each one individually?

src/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -25,7 +25,14 @@ export function createDirectAnswer(data: any): FeaturedSnippetDirectAnswer | Fie
...
     const fieldType = commonDirectAnswerData.fieldType;
-    if (fieldType != BuiltInFieldType.MultiLineText && fieldType != BuiltInFieldType.RichText) {
+    const supportedFieldTypes: BuiltInFieldType[] = [
+      BuiltInFieldType.MultiLineText,
+      BuiltInFieldType.RichText,
+      BuiltInFieldType.RichText_v2,
+      BuiltInFieldType.Markdown,
+      BuiltInFieldType.Html
+    ];
+    if (!supportedFieldTypes.includes(fieldType)) {
```
benmcginnis, line 35: Is this error actually helpful for end-users and implementers? It seems like a type check on the API response which I'm not clear on the value of

saahithjanapati: We thought it might be useful in case new field types get implemented for the API, but aren't added here. I don't think end-users should be seeing that often if this is kept up to date.

In the latest commit, we changed the method to return undefined if the type isn't supported (so it doesn't crash the application) and changed the error to a console warning.

src/transformers/searchservice/createDirectAnswer.ts
```diff
@@ -25,8 +28,16 @@ export function createDirectAnswer(data: any): FeaturedSnippetDirectAnswer | Fie
     };
   } else if (isFeaturedSnippetDirectAnswer) {
     const fieldType = commonDirectAnswerData.fieldType;
-    if (fieldType != BuiltInFieldType.MultiLineText && fieldType != BuiltInFieldType.RichText) {
-      throw new Error(`Unexpected fieldType for featured snippet direct answer: ${fieldType}`);
+    const supportedFieldTypes: BuiltInFieldType[] = [
```
benmcginnis, line 31: this is a super minor nit, but this constant won't change on each invocation of the function so it should probably be defined out of the scope of the function

saahithjanapati: I just moved it outside the createDirectAnswer() method in the latest commit

benmcginnis (approved):

PR #248, opened 2023-09-13
Update buildchain to use rollup in order to improve esm support

cea2aj:
The library exported an esm version, however it didn't work with all esm tools becase some require "type": "module" to be set, or the files to end in ".mjs". By moving over to rollup, we can output the esm files with the ".mjs" extension which makes this library work with vite.

This change is based on the approach which Chat decided on: https://github.com/yext/chat-ui-react/pull/38

J=BACK-2527
TEST=manual

Manually ran the test site locally and confirmed the requests still fire correctly. I still need to test this with vite.


cea2aj:
Upon further testing, the existing library works just fine with the pages js dev and prod builds, so I don't think this change is needed

PR #259, opened 2024-01-25
Resolve Vulnerabilities

EmilyZhang777:
This PR addresses the following vulnerabilities:
1. api key being exposed
2. insecure document methods
3. webpack version

J=VULN-37165,VULN-3766,VULN-37733,VULN-37734,VULN-38429,VULN-38430
TEST=auto

Ran the following scripts and made sure they ran successfully.
`npm run dev`, `npm run test`, `npm run dev` in test-site

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "jest-environment-jsdom": "^28.1.0",
     "ts-loader": "^8.4.0",
     "typescript": "^4.0.3",
-    "webpack": "^5.73.0",
+    "webpack": "^5.76.0",
     "webpack-cli": "^4.4.0"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "jest-environment-jsdom": "^28.1.0",
     "ts-loader": "^8.4.0",
     "typescript": "^4.0.3",
-    "webpack": "5.90.0",
+    "webpack": "^5.90.0",
     "webpack-cli": "^4.4.0"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


benmcginnis (approved):

nmanu1 (commented):
I didn't see it in the testing details in your PR description, so just to double check, you went to both the JS and TS test sites and saw no changes in behavior after this change?

EmilyZhang777:
Made sure the test sites look the same as before 👍

benmcginnis (approved):

test-site/webpack.config.js
```diff
@@ -20,6 +21,11 @@ module.exports = [{
     path: path.resolve(__dirname, 'dist'),
     libraryTarget: 'window'
   },
+  plugins: [
+    new webpack.DefinePlugin({
```
nmanu1, line 25: nit: can this be assigned to a variable and then used here and below so the logic isn't repeated?

package.json
```diff
@@ -74,7 +74,7 @@
     "jest-environment-jsdom": "^28.1.0",
     "ts-loader": "^8.4.0",
     "typescript": "^4.0.3",
-    "webpack": "^5.73.0",
+    "webpack": "^5.76.0",
```
nmanu1, line 77: since based on the package-lock, it looks like we've upgraded to v5.90.0, can we update this value to match that? it makes it easier to keep track

test-site/package.json
```diff
@@ -6,7 +6,7 @@
   "devDependencies": {
     "ts-loader": "^9.2.6",
     "typescript": "^4.4.3",
-    "webpack": "^5.58.1",
+    "webpack": "5.76.0",
```
nmanu1, line 9: can we pin to the major version again instead of to this patch?

nmanu1 (approved):

PR #263, opened 2024-06-04
Add generativeDirectAnswer service

anguyen-yext2:
J=CLIP-1226
TEST=auto,manual

ran and saw jest test pass.
tested the gda endpoint locally via core test site, saw appropriate api resp
onse.


EmilyZhang777 (commented):
LGTM! But will defer to Clippy. Only have a nit

src/validation/ApiResponseValidator.ts
```diff
@@ -33,6 +33,7 @@ export class ApiResponseValidator {
 
   private checkForApiErrors(apiResponse: ApiResponse): SearchError | undefined {
     if (apiResponse.meta?.errors?.length >= 1) {
+      apiResponse.meta.errors.map(e => console.log(e));
```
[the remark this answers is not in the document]
anguyen-yext2: oh oops. I forgot to remove it. Thanks for pointing it out!

src/models/generativedirectanswer/GenerativeDirectAnswerRequest.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @public
+ */
+export interface GenerativeDirectAnswerRequest extends SearchRequest {
+  /** The ID of the search request. */
+  searchId: string,
+  /** The text of the user-written query that prompted Search results. */
+  searchTerm: string,
+  /** The complete set of Search Results */
+  results: VerticalResults | Record<string, VerticalResults[]>
```
yen-tt, line 15: perhaps I missed something, but why is the type a map of key to results? why not just an array of VerticalResults?

anguyen-yext2: we defined results in [GDAApiRequestBody](https://gerrit.yext.com/plugins/gitiles/alpha/+/refs/heads/master/src/com/yext/search/generativeanswers/generative_answers.proto#18) to be of type struct, so we can't pass in array. `VerticalResults` is for a Vertical Search Result, `Record<string, VerticalResults[]>` is for a Universal Search Result. 
I think there is another way I could accomplish this: keep results to be of type `VerticalResults[]`, then in `GDAServiceImpl`, convert request.results to object before sending the api request. Something like this:
```
let results: Object = {}
if (request.results.length === 1) {
    results = request.results[1]
} else if (request.results.length > 1) {
    results = {verticals: request.results}
}
```
let me know if you prefer this way, then I can change and test it

yen-tt: yeah, I think that would work better so user of core don't have to transform the universal results themselves

src/infra/GenerativeDirectAnswerServiceImpl.ts
```diff
@@ -51,10 +51,16 @@ export class GenerativeDirectAnswerServiceImpl implements GenerativeDirectAnswer
       visitorIdMethod: this.config.visitor?.idMethod,
       ...this.config?.additionalQueryParams
     };
+    let results = {};
```
yen-tt, line 54: nit: add type

yen-tt (commented):
are you going to publish a new version or wait for search's publish cycle? If it's the former, then you should bump the version in package.json

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-core",
-  "version": "2.5.2",
+  "version": "2.6.0.beta",
```
yen-tt, line 3: super nit: can we follow the past version format from Search and use hyphen instead? ([example](https://www.npmjs.com/package/@yext/chat-ui-react?activeTab=versions)L 2.6.0-beta)

PR #266, opened 2024-06-21
make `queryDurationMillis` field optional in VerticalResults

anguyen-yext2:
J=CLIP-1318
TEST=auto,manual

ran `npm run test`. Tested this change together with the rest of the sdk changes to confirm that gda is executed successfully after executeVerticalQuery finishes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-core",
-  "version": "2.6.0-beta",
+  "version": "2.6.0-beta.2",
   "description": "Typescript Networking Library for the Yext Search API",
   "main": "./dist/commonjs/src/index.js",
   "module": "./dist/esm/src/index.js",
--- a/src/models/searchservice/response/VerticalResults.ts
+++ b/src/models/searchservice/response/VerticalResults.ts
@@ -11,7 +11,7 @@ export interface VerticalResults {
   /** A array of {@link AppliedQueryFilter}s which were applied to the vertical results. */
   appliedQueryFilters: AppliedQueryFilter[],
   /** The duration of the query in milliseconds. */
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-core",
-  "version": "2.6.0-beta",
+  "version": "2.6.0-beta.2",
```
Fondryext, line 3: silly question - is it normal/typical to go to .2 instead of .1? I'm not sure how versioning works

src/models/searchservice/response/VerticalResults.ts
```diff
@@ -11,7 +11,7 @@ export interface VerticalResults {
   /** A array of {@link AppliedQueryFilter}s which were applied to the vertical results. */
   appliedQueryFilters: AppliedQueryFilter[],
   /** The duration of the query in milliseconds. */
-  queryDurationMillis: number,
+  queryDurationMillis?: number,
```
yen-tt, line 14: can you specify in the comment when it is present / not present?

package.json
anguyen-yext2: I'm not sure neither, but I was told to go from x.x.x-beta to x.x.x-beta.2
I have already merged the beta.2 version for search-headless before, so I assume this is ok?

vijay267 (approved):

package.json
vijay267: I had the same question but I'm not sure either. Maybe Yen knows? If you've done it before, perhaps it's ok.

anguyen-yext2: Yen did told me that there is no strict rule on the revision numbers on beta. She personally view the original as the first version so next up should be .2

PR #268, opened 2024-07-12
search-core: Adding Page Number to Segment

vdarosa:
Added an optional parameter called pageNumber to Segments.ts

Test: Existing tests in search-core all pass


docs/search-core.segment.pagenumber.md
```diff
@@ -0,0 +1,13 @@
+<!-- Do not edit this file. It is automatically generated by API Documenter. -->
+
+[Home](./index.md) &gt; [@yext/search-core](./search-core.md) &gt; [Segment](./search-core.segment.md) &gt; [pageNumber](./search-core.segment.pagenumber.md)
+
+## Segment.pageNumber property
+
+The page number of a segment from a pdf
```
EmilyZhang777, line 7: I'd prefer to say something like "The page number of the document that the segment came from. Only populated when document is a PDF file."
If you end up changing this, make sure to update other places as well

EmilyZhang777:
Also, this pr needs to be merged into a hotfix branch instead of master

EmilyZhang777 (approved):

PR #271, opened 2024-08-14
search-core: add support for FunctionVertical

alapre:
This PR adds a Source enum type for Function Vertical, updates ResultFactory.ts to handle a function vertical response, and adds auto test.

Test: manual, auto
Tested with auto test and through test site, ensuring function vertical data appear in the response.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-core",
-  "version": "2.6.0-beta.2",
+  "version": "2.6.0-beta.3",
   "description": "Typescript Networking Library for the Yext Search API",
   "main": "./dist/commonjs/src/index.js",
   "module": "./dist/esm/src/index.js",
--- a/src/transformers/searchservice/ResultsFactory.ts
+++ b/src/transformers/searchservice/ResultsFactory.ts
@@ -26,6 +26,8 @@ export class ResultsFactory {
           return this.fromAlgoliaSearchEngine(result, resultIndex);
         case Source.DocumentVertical:
           return this.fromDocumentVertical(result, resultIndex);
+        case Source.FunctionVertical:
+          return this.fromFunctionVertical(result, resultIndex);
         default:
           return this.fromCustomSource(result, resultIndex);
       }
@@ -135,6 +137,19 @@ export class ResultsFactory {
       link: result.website,
       id: result.id,
       entityType: result.type,
+    };
+  }
+
+  public static fromFunctionVertical(result: any, index: number): Result {
+    const rawData = result.data ?? result;
+    return {
+      rawData: rawData,
+      source: Source.FunctionVertical,
+      index: index,
+      name: rawData.name,
+      description: rawData.description,
+      link: rawData.website,
+      id: rawData.id,
     };
   }
 }
--- a/src/transformers/searchservice/ResultsFactory.ts
+++ b/src/transformers/searchservice/ResultsFactory.ts
@@ -115,10 +115,18 @@ export class ResultsFactory {
   }
 
   private static fromCustomSource(result: any, index: number): Result {
+    return this.fromCustomSourceHelper(result, index, Source.Custom);
+  }
+
+  private static fromFunctionVertical(result: any, index: number): Result {
+    return this.fromCustomSourceHelper(result, index, Source.FunctionVertical);
+  }
+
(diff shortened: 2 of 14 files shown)


Fondryext (commented):
So quick feedback about process:
We do eventually want this change into the develop branch, but we'll have to go through a few steps to get it there. This wiki page is a good guide: https://wiki.yext.to/index.php/Public_SDK_release_process

I think for your change we'll probably want to have this be a new minor version, 2.6.0, but we can confirm with Baigel really quickly, and/or see if chat is currently doing any generative direct answer work in this repo that might complicate things. 

If we do decide to make this 2.6.0, then we'll make a new branch, called maybe release/v2.6, and then you'll make a PR from your dev/update-... branch into that one. When that's merged we'll then merge release/ into master, and then finally can merge master/ into develop.

src/transformers/searchservice/ResultsFactory.ts
```diff
@@ -137,4 +139,17 @@ export class ResultsFactory {
       entityType: result.type,
     };
   }
+
+  public static fromFunctionVertical(result: any, index: number): Result {
```
vijay267, line 143: Can we put this next to "fromCustomSource"? Plus we probably could refactor the two methods to use a helper since the only difference if the source between the two of them :)

test-site/src/ts/index.ts
```diff
@@ -8,7 +8,7 @@ import initDirectAnswers from './initDirectAnswers';
 
 const coreConfig: SearchConfig = {
   apiKey: process.env.API_KEY,
-  experienceKey: 'slanswers',
+  experienceKey: 'developer-support-test',
```
vijay267, line 11: Does this experience key impact all of the test site options? Like universal search, vertical search, etc.? If so, let's make sure everything still works

vijay267 (commented):
Also are the doc files generated? Wondering why it decided to change the markdown format now...

test-site/src/ts/index.ts
alapre: I adjusted this experience to make sure everything still works for the ts page. 
Flagging that for the js page, in index.js, the experience key is 'rosetest' which I think no longer exists (at least isn't an experience for the two businesses with experiences mentioned in [this](https://docs.google.com/document/d/1EusosNnvHpT_J3RvTnz1IjnQBGQ_unnI7tfBJFoRzQA/edit?usp=sharing) doc)? So the js test site doesn't really work (says `invalid params. experienceKey: 'rosetest'` when you try most of the searches).

vijay267: I'm guessing we archived the rosetest experience at some point.

src/transformers/searchservice/ResultsFactory.ts
```diff
@@ -115,10 +115,18 @@ export class ResultsFactory {
...
 
   private static fromCustomSource(result: any, index: number): Result {
+    return this.fromCustomSourceHelper(result, index, Source.Custom);
+  }
+
+  private static fromFunctionVertical(result: any, index: number): Result {
+    return this.fromCustomSourceHelper(result, index, Source.FunctionVertical);
+  }
+
+  private static fromCustomSourceHelper(result: any, index: number, source: Source): Result {
```
vijay267, line 125: Maybe name this like "fromCustomDataHelper"? CustomSource works well for "fromCustomSource", but sounds a bit odd for "fromFunctionVertical"

test-site/src/ts/index.ts
alapre: Looks like for one of the checks (run linter), 'rosetest' is needed in order for it to pass - leaving it as is for now but can look into this more!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-core",
-  "version": "2.6.0-beta.2",
+  "version": "2.6.0-beta.3",
```
Fondryext, line 3: ah, one last little 'gotcha' with these github changes I forgot to mention (sorry!)
Now that we've changed the version in `package.json`, if you run `npm i` it will update the autogenerated `package-lock.json` and update the version there as well. Could you do that really quick + add it to the PR? The change looks good otherwise!

Fondryext (approved):
LGTM! Maybe double-check with Deepak about that one conversation left, but I think it's fine

PR #281, opened 2024-11-12
Use Cached Endpoints

mkouzel-yext:
This change updated the LiveAPI endpoints that search-core provides so that it provides the cached version of the endpoints. Using the cached endpoints helps improve performance.

J=WAT-4442
TEST=manual

Built test site with new change, saw that the cached endpoints were being used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-core",
-  "version": "2.5.4",
+  "version": "2.5.5",
   "description": "Typescript Networking Library for the Yext Search API",
   "main": "./dist/commonjs/src/index.js",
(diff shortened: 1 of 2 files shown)


Fondryext:
So the way we usually make changes to our github repos is:
1. Make a new branch for the new version. In this case this'll be 2.5.5, so the branch will probably be called `hotfix/v2.5.5`
2. Up the package version on that branch correspondingly. In a lot of our repos this change gets created automatically
3. Make a pull request to merge your changes into the hotfix branch
4. Once all changes are in the hotfix branch, *then* make a request from the hoftix branch into master
5. Once that's pulled, we make a new release tag and publish to npm, and possibly pull into develop

Happy to expand on any of these

PR #282, opened 2024-11-12
Use Cached Endpoints

mkouzel-yext:
This change updated the LiveAPI endpoints that search-core provides so that it provides the cached version of the endpoints. Using the cached endpoints helps improve performance.

J=WAT-4442
TEST=manual

Built test site with new change, saw that the cached endpoints were being used.


Fondryext:
In this case the test failures are real, provideEndpointsTest.ts is now failing. Change looks good otherwise though

Fondryext (approved):

PR #283, opened 2024-11-12
Hotfix/v2.5.5

mkouzel-yext:
Changes the endpoint for autocompletes to the cached version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yext/search-core",
-  "version": "2.5.4",
+  "version": "2.5.5",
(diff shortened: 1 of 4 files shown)


k-gerner (approved):

Fondryext (changes requested):
Two thoughts - one, I would expect package-lock.json to have also been updated. Could you make sure `npm i` is run, so that file is updated as well?

Second, when it's time to merge in to master we're usually a little more standardized with our commit messages. Here is an example: https://github.com/yext/search-core/pull/278

Fondryext (approved):

PR #287, opened 2024-11-22
search-core: Resolve Vulnerabilities

mkouzel-yext:
Regenerated package-lock.json and test-site/package-lock.json to ensure that both used updated safe dependencies.

J=VULN-39421
TEST=manual

Inspected babel dependencies in each file to confirm they used safe versions.


Fondryext:
A bit vague of an ask but could we try and see why all the .md files got reformatted? I wouldn't expect anything in this to have done it since we just updated package-lock. If it's from an earlier package change and only just got triggered, that's fine. Don't spend much time on this, just a quick glance at previous package changes to see if any seem related

mkouzel-yext:
I don't see anything obvious that caused the docs to get reformatted. I checked my local commit, and I did not include anything in the `docs` directory locally so it was done automatically.

Fondryext (approved):
no worries, then. Here's hoping they stay formatted for a while lol I feel like it happens every couple package upgrades

PR #304, opened 2025-07-29
chore: mention max length for visitor idmethod in docs

k-gerner:
Updating the docs to mention that the max length of the visitor idmethod is 16 (taken from Snowflake table). Not planning on making a new release for this, and will instead wait for the next release to pick up this change.

### DIFF
--- a/src/models/core/Visitor.ts
+++ b/src/models/core/Visitor.ts
@@ -9,7 +9,7 @@
 export interface Visitor {
   /** The ID associated with the user */
   id: string,
-  /** The type of visitor
+  /** The type of visitor. Max of 16 characters.
    *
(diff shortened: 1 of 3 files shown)


src/models/core/Visitor.ts
```diff
@@ -9,7 +9,7 @@
 export interface Visitor {
   /** The ID associated with the user */
```
vijay267, line 10: I realize I didn't mention this in the item, but it'd be good to also mention that the visitor id (not the method) has a limit of 64 characters.

k-gerner: https://github.com/yext/search-core/pull/306

